PR #19, opened 2018-09-10
cmd/telemeter-server: register handler request size metrics

s-urbaniak:
This helps us with back-of-the-envelope calculating the expected
request sizes.

/cc @smarterclayton @brancz

### DIFF
--- a/cmd/telemeter-server/main.go
+++ b/cmd/telemeter-server/main.go
@@ -79,6 +79,10 @@ var (
 		Name: "telemeter_server_request_federate",
 		Help: "Tracks the number of federated uploads.",
 	}, []string{"code"})
+	metricRequestFederateSize = prometheus.NewSummaryVec(prometheus.SummaryOpts{
+		Name: "telemeter_server_request_federate_size_bytes",
+		Help: "Tracks the sum of request sizes of metrics received",
+	}, []string{"code"})
 	metricRequestFederateLatency = prometheus.NewSummaryVec(prometheus.SummaryOpts{
 		Name: "telemeter_server_request_federate_latency",
 		Help: "Tracks latency on federated uploads.",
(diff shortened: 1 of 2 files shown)


cmd/telemeter-server/main.go
```diff
@@ -79,14 +79,18 @@ var (
 		Name: "telemeter_server_request_federate",
 		Help: "Tracks the number of federated uploads.",
 	}, []string{"code"})
+	metricRequestFederateSize = prometheus.NewSummaryVec(prometheus.SummaryOpts{
```
brancz: The naming of these confused me, how about we actually just call them after the endpoints name "upload". Federate suggests to me, that it has something to do with the `/federate` endpoint.

squat: I think it was for consistency; the other metrics that deal with the `/upload` endpoint, e.g. `metricRequestFederate` and `metricRequestFederateLatency`, also use `Federate` in the name. I agree that it is confusing since in Prometheus world `federation` has a very specific use and meaning. We should keep it consistent but consider renaming them all

brancz: These are actually just normal http metrics, how about having the endpoint name as a label? All http metrics in Prometheus are instrumented that way. Then it's easier to use the same aggregation rules for different endpoints, which would make sense.

s-urbaniak: ack, I was mostly following existing codebase conventions here, in order not to introduce too much noise and changes.

More than happy though to do a more invasive change here.

brancz: Shouldn't be a lot more invasive than the current changes, let's do them and make a good example for instrumentation :slightly_smiling_face: .

s-urbaniak: @brancz @squat :-( I am having tremendous troubles bumping this.
The used prom client library is pretty old, so `MustCurryWith...` is not available, so I cannot introduce nice reusable metrics objects. Bumping it doesn't work right now, as parts of github is down, so `glide up` doesn't work right off the bat.

Hence: I suggest to add the suggested changes as a follow-up PR. It's not super nice, but at least we would have some metrics and need not to wait for github to be up again.

brancz: sounds good to me given the current constraints

PR #20, opened 2018-09-12
Rename metrics

squat:
cc @s-urbaniak @brancz 

This PR renames the metrics exposed by the telemeter server. It makes the http metrics generic and uses labels to create different a timeseries for each endpoint, following the Prometheus best practices.


cmd/telemeter-server/main.go
```diff
@@ -468,3 +454,17 @@ func loadPrivateKey(data []byte) (interface{}, error) {
 
 	return nil, fmt.Errorf("unable to parse private key data: '%s', '%s' and '%s'", err0, err1, err2)
 }
+
+// instrumentHandler instruments an http.HandlerFunc.
+func instrumentHandler(handlerName string, handler http.HandlerFunc) http.HandlerFunc {
```
s-urbaniak: :+1:

squat:
@s-urbaniak any idea why Prow still reports the test job as not finished? Looking at the logs, the build succeeded and then Prow has logged the line `2018/09/12 12:35:10 Pod release-latest already succeeded in 16s` ~2000 times

PR #21, opened 2018-09-17
pkg/transform: rename to metricfamily, separate source files

s-urbaniak:
This is just a mechanical refactor for better readability.


pkg/metricfamily/sort.go
```diff
@@ -0,0 +1,131 @@
...
+
+	clientmodel "github.com/prometheus/client_model/go"
+)
+
+func SortMetrics(family *clientmodel.MetricFamily) (bool, error) {
+	sort.Sort(MetricsByTimestamp(family.Metric))
+	return true, nil
+}
+
+type MetricsByTimestamp []*clientmodel.Metric
```
metalmatze: Do we really need to use these "old" interfaces? I think you can simply use the `sort.Slice` func with the `func (m MetricsByTimestamp) Less(i int, j int) bool` as parameter. 😊

pkg/metricfamily/sort.go
```diff
@@ -0,0 +1,131 @@
...
+		dst.Metric = result
+		families[pos] = nil
+	}
+	return Pack(families)
+}
+
+// PackedFamilyWithTimestampsByName sorts a packed slice of metrics
+// (no nils, all families have at least one metric, and all metrics
+// have timestamps) in order of metric name and then oldest sample
+type PackedFamilyWithTimestampsByName []*clientmodel.MetricFamily
```
metalmatze: I can't find a reference of this in the diff anymore. 🤔

s-urbaniak: It was here: https://github.com/openshift/telemeter/blob/676f4b315e50419457013d8180a6bc20f8be74db/pkg/transform/transform.go#L39

pkg/metricfamily/sort.go
s-urbaniak: Agreed :+1: This is just a first mechanical refactor (hopefully) without actual logic changes.

squat: right! logic and opinionated clean up will follow shortly!

PR #24, opened 2018-09-21
pkg/metricsclient: disable keep alives

s-urbaniak:
Currently, telemeter-client leaves the connection open between subsequent pushes.
This quickly exhausts file descriptors on server-side.

As telemeter client metric pushes are far apart (4.5 minutes), it doesn't make sense
to enable this setting on the client.

/cc @squat @brancz: This was the cause for yesterday's benchmark failures.
The server-side probably had a too low open files ulimit, causing new connections to be rejected.
With this tweak, I can launch 10000 local mock clients easily and send metrics consistently.

### DIFF
--- a/pkg/metricsclient/metricsclient.go
+++ b/pkg/metricsclient/metricsclient.go
@@ -226,5 +226,6 @@ func DefaultTransport() *http.Transport {
 			KeepAlive: 30 * time.Second,
 		}).Dial,
 		TLSHandshakeTimeout: 10 * time.Second,
+		DisableKeepAlives:   true,
 	}
 }


s-urbaniak:
I wonder if it makes also sense to disable keep alives for the authorize endpoint, as I was able to get into saturation. I guess, this is fine though as this potentially can also grow without bounds:

https://github.com/openshift/telemeter/blob/06fdf0564645202159eb7bffce1b85594f071b63/cmd/telemeter-server/main.go#L200-L207

We should have an eye (metrics) during benchmarks of all telemeter connections, currently only telemeter-to-telemeter and telemeter-to-authorization communication.

brancz:
Awesome find!

PR #25, opened 2018-09-25
init jsonnet and create telemeter-client libsonnet

squat:
This pull request creates the basic elements necessary to package and deploy a Telemeter client instance. The manifests can be customized and generated using jsonnet. In order to build the manifests, run `make`.

Follow up PRs will:
* add a `jsonnet/telemeter/server` directory with the jsonnet needed to generate the Telemeter server manifests
* add `jsonnet/telemeter/*/{alerts,rules,dashboards}` directories to implement the monitoring mixin interface
* remove the `deploy` directory, which will be fully superseded by the new `manifests` directory

cc @s-urbaniak @brancz


manifests/telemeter-client-deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+        k8s-app: telemeter-client
+    spec:
+      containers:
+      - command:
+        - /usr/bin/telemeter-client
+        - --id=$(ID)
+        - --from=https://prometheus-k8s.openshift-monitoring.svc:9091
+        - --from-ca-file=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
+        - --from-token-file=/var/run/secrets/kubernetes.io/serviceaccount/token
+        - --to=
```
s-urbaniak: is this a left-over?

jsonnet/telemeter/client/kubernetes/kubernetes.libsonnet
```diff
@@ -0,0 +1,112 @@
...
+
+      local telemeterClient =
+        container.new('telemeter-client', $._config.imageRepos.telemeterClient + ':' + $._config.versions.telemeterClient) +
+        container.withCommand([
+          '/usr/bin/telemeter-client',
+          '--id=$(ID)',
+          '--from=' + $._config.telemeterClient.from,
+          '--from-ca-file=' + fromCAFile,
+          '--from-token-file=' + fromTokenFile,
+          '--to=' + $._config.telemeterClient.to,
```
s-urbaniak: ah, I see, this is coming from configuration :+1:

jsonnet/main.jsonnet
```diff
@@ -0,0 +1,7 @@
+local t = (import 'telemeter-client/telemeter-client.libsonnet') + {
+  _config+:: {
+    namespace: 'openshift-monitoring',
```
brancz: this can just be the default config, no need to specify it here I think

jsonnet/telemeter/client/kubernetes/kubernetes.libsonnet
```diff
@@ -0,0 +1,112 @@
...
+local credentialsSecret = 'telemeter-client';
+local credentialsVolumeName = 'credentials';
+local credentialsMountPath = '/etc/telemeter';
+local fromCAFile = '/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt';
+local fromTokenFile = '/var/run/secrets/kubernetes.io/serviceaccount/token';
+local metricsPort = 8080;
+
+{
+  _config+:: {
+    namespace: 'default',
```
brancz: this should be `openshift-monitoring`

jsonnet/telemeter/client/kubernetes/kubernetes.libsonnet
```diff
@@ -0,0 +1,112 @@
...
+
+    telemeterClient+:: {
+      from: 'https://prometheus-k8s.openshift-monitoring.svc:9091',
```
brancz: the namespace should be templated here

jsonnet/telemeter/client/telemeter-client.libsonnet
```diff
@@ -0,0 +1,11 @@
+(import 'kubernetes/kubernetes.libsonnet') + {
+  _config+:: {
+    namespace: 'default',
+
+    telemeterClientSelector: 'job="telemeter-client"',
+
+    jobs: {
+      TelemeterClient: $._config.telemeterClientSelector,
```
brancz: TelemeterClient -> telemeterClient

jsonnet/telemeter/client/telemeter-client.libsonnet
```diff
@@ -0,0 +1,11 @@
+(import 'kubernetes/kubernetes.libsonnet') + {
+  _config+:: {
+    namespace: 'default',
+
+    telemeterClientSelector: 'job="telemeter-client"',
```
brancz: as long as we have everything under control for this I don't think we need to make this configurable

jsonnet/telemeter/client/telemeter-client.libsonnet
```diff
@@ -0,0 +1,11 @@
+(import 'kubernetes/kubernetes.libsonnet') + {
+  _config+:: {
+    namespace: 'default',
```
brancz: the namespace is already defined in the kubernetes/kubernetes.libsonnet file, so this doesn't need to be here, let's strive to keep configuration local to its use

s-urbaniak:
(ok, still getting used to prow)

jsonnet/telemeter/client/kubernetes/kubernetes.libsonnet
squat: yes, once we know the production telemeter URL we should use that as the default value in this libsonnet file

jsonnet/telemeter/client/telemeter-client.libsonnet
squat: I was following the convention in https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus.libsonnet#L38-L46

should they all be lowercase?

squat:
I'll post a follow up addressing @brancz 's comments :+1:

jsonnet/telemeter/client/telemeter-client.libsonnet
squat: :+1:

jsonnet/telemeter/client/kubernetes/kubernetes.libsonnet
squat: very good call. the current way, users had to update two values, with this change they will only need to change the namespace.

jsonnet/telemeter/client/telemeter-client.libsonnet
squat: got it. I was trying to follow the conventions in kube-prometheus but think about this detail

jsonnet/telemeter/client/telemeter-client.libsonnet
brancz: that's strange, I don't know why we ever started doing that, we should change that in the kubernetes mixin as well

brancz: good catch!

PR #27, opened 2018-09-26
pkg/authorizer: adapt to tollbooth API changes

s-urbaniak:
This is a first best efforts refactoring to adapt to Tollbooth API changes.

/cc @squat


pkg/authorizer/server/server.go
```diff
@@ -79,7 +79,7 @@ func (a *Authorizer) AuthorizeHTTP(w http.ResponseWriter, req *http.Request) {
 	userToken := auth[1]
 
 	var (
-		resp *TokenResponse
+		resp *ClusterRegistration
 		err  error
 	)
 	if a.to == nil {
```
squat: this adhoc polymorphism is a little funny. it would be cleaner to configure the Authorizer struct with an authorize func rather than switching different funcs based on the `to` field. we can clean this up in a followup

pkg/authorizer/server/types.go
```diff
@@ -1,21 +1,12 @@
 package server
 
-type TokenRequest struct {
-	APIVersion string `json:"api_version"`
-
-	AuthorizationToken string `json:"authorization_token"`
+type ClusterRegistration struct {
```
squat: it looks like this type is never used outside of this package. can we make it unexported?

pkg/authorizer/server/types.go
```diff
@@ -1,21 +1,12 @@
...
-
-	Status  string `json:"status"`
-	Code    int    `json:"code"`
-	Reason  string `json:"reason"`
-	Message string `json:"message"`
-
-	AccountID string `json:"account_id"`
-
-	Labels map[string]string `json:"labels"`
+type Error struct {
```
squat: Similarly, this type is never used outside of this package. As long as the fields are exported, an unexported type can still be marshaled to JSON.

pkg/authorizer/server/server.go
s-urbaniak: yes, totally agreed, I wanted this factoring to be least invasive, merely reflecting necessary changes towards tollbooth, otherwise the review would have been messier.

I have also many ideas how to move forward with this source code base and we definitely should move forward with this once we have the interface with tollbooth straight.

pkg/authorizer/server/types.go
s-urbaniak: yes, we can definitely make this unexported. In the future this should also get proper godoc, or even be auto-generated from the OpenAPI spec.

s-urbaniak:
@squat PTAL

PR #28, opened 2018-09-26
jsonnet: put telemeter server URL into secret and env var

squat:
This PR puts the telemeter server URL into a secret and loads it into the container using an environment variable. This has the benefit of being easier to override in actual deployments by simple modifying the underlying secret rather than needing to iterate over the flags and modifying the one with the URL.

This helps address: https://github.com/openshift/cluster-monitoring-operator/pull/103#discussion_r220598831

cc @s-urbaniak @brancz


brancz:
Nice! I like this solution!

PR #30, opened 2018-09-27
makes telemeter-server.yaml template

riuvshin:
Template will allow us to override image / image tag

### DIFF
--- a/deploy/telemeter-server.yaml
+++ b/deploy/telemeter-server.yaml
@@ -6,7 +6,7 @@
 apiVersion: v1
 kind: Template
 metadata:
-  name: openshiftio-tenant-log-indirector-service
+  name: telemeter
 objects:
 - apiVersion: v1
   kind: Service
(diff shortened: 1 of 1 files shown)


s-urbaniak:
good stuff, thanks @riuvshin !

deploy/telemeter-server.yaml
```diff
@@ -104,3 +107,8 @@ items:
           - --tls-key=/etc/pki/service/tls.key
           - --tls-crt=/etc/pki/service/tls.crt
           - --authorize=http://telemeter-authorization
+parameters:
+- name: IMAGE
+  value: openshift/origin-telemeter
+- name: IMAGE_TAG
+  value: v3.11
```
s-urbaniak: post-mortem comment: this should be `v4.0`, no?

squat: I think we should specify that in the saasherder config https://github.com/app-sre/saas-telemeter/blob/master/telemeter-services/telemeter-server.yaml#L10.

@riuvshin kept it this way to be identical to the original manifest. 

We can make a follow up PR and update the image to quay.io/openshift/origin-telemeter and the tag to v4.0.0 but it's independent of this change

riuvshin: exactly, thanks @squat :)

PR #35, opened 2018-09-28
add service account with correct roles to telemeter-client

squat:
In order to authenticate against a Prometheus k8s pod created by the Cluster Monitoring Operator, the bearer token must belong to a ServiceAccount with the following SAR `{"resource":"namespaces","verb":"get"}` [0]

[0] https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/prometheus.jsonnet#L290

cc @s-urbaniak

### DIFF
--- a/jsonnet/jsonnetfile.lock.json
+++ b/jsonnet/jsonnetfile.lock.json
@@ -8,7 +8,7 @@
                     "subdir": "jsonnet/telemeter/client"
                 }
             },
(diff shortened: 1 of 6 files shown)


brancz:
For consistency I think it would make sense to rather bind to the cluster-monitoring-view Role, because if we change that one the telemeter ClusterRole now would have to be adapted as well.

squat:
@brancz great idea. I'll test this and make a new pr

PR #36, opened 2018-10-01
secure telemeter client with rbac proxy

squat:
Currently, nothing can scrape the Telemeter clients metrics because it is only listening on localhost. In order to expose the Telemeter client's metrics to a prometheus server, we must allow other pods to access its `/metrics` endpoint. However, the Telemeter client can potentially expose sensitive user data on its metrics endpoint. For this reason, we need to add a proxy that exposes the metrics endpoint only to authorized applications.

cc @s-urbaniak

### DIFF
--- a/jsonnet/telemeter/client/kubernetes/kubernetes.libsonnet
+++ b/jsonnet/telemeter/client/kubernetes/kubernetes.libsonnet
@@ -2,9 +2,13 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
 local credentialsSecret = 'telemeter-client';
 local credentialsVolumeName = 'credentials';
 local credentialsMountPath = '/etc/telemeter';
+local tlsSecret = 'telemeter-client-tls';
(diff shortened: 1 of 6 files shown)


jsonnet/telemeter/client/kubernetes/kubernetes.libsonnet
```diff
@@ -39,17 +46,31 @@ local metricsPort = 8080;
...
+                                  'subjectaccessreviews',
+                                ]) +
+                                policyRule.withVerbs(['create']);
+
+      local coreRule = policyRule.new() +
+                       policyRule.withApiGroups(['']) +
+                       policyRule.withResources([
+                         'namespaces',
+                       ]) +
+                       policyRule.withVerbs(['get']);
```
s-urbaniak: just a very small nit (can be done as a follow-up by me too): we should document this somewhere, as the "whoever can get namespaces, can scrape me" idiom seems to be ubiquitous.

PR #37, opened 2018-10-01
add match file to Telemeter client

squat:
This PR adds the `match-file` flag to the Telemeter client deployment.

cc @s-urbaniak

### DIFF
--- a/jsonnet/telemeter/client/kubernetes/kubernetes.libsonnet
+++ b/jsonnet/telemeter/client/kubernetes/kubernetes.libsonnet
@@ -1,7 +1,8 @@
 local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
-local credentialsSecret = 'telemeter-client';
-local credentialsVolumeName = 'credentials';
-local credentialsMountPath = '/etc/telemeter';
+local secretName = 'telemeter-client';
+local secretVolumeName = 'secret-telemeter-client';
+local secretMountPath = '/etc/telemeter';
+local matchFileName = 'match-rules';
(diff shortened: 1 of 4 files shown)


jsonnet/telemeter/client/kubernetes/kubernetes.libsonnet
```diff
@@ -18,6 +19,14 @@ local securePort = 8443;
       from: 'https://prometheus-k8s.%(namespace)s.svc:9091' % $._config,
       serverName: 'server-name-replaced-at-runtime',
       to: '',
+      matchRules: [
```
brancz: Just a suggestion, not important at all right now: how about we extract this into a json configuration file, I think that might be clearer to teams to contribute their matchers, as opposed to them having to have the mental overhead of the surrounding jsonnet code.

squat: agreed, this would allow for improved:
1. introspection, this way it is easier to even answer the question "what metrics are being collected"
2. contribution

Ideally, this should live in the root of the repo so it is easiest to introspect. The only issue with this is that `jsonnet/telemeter/client/jsonnetfile.json` cannot declare the dependency as:
```json
        {
            "name": "metrics",
            "source": {
                "git": {
                    "remote": "../../../",
                    "subdir": ""
                }
            },
            "version": "."
        }
```
since this is a library itself and `../../../` is relative to the directory where `jb` is invoked. So if we want the metrics at the root, which I think would be good, we have to add an explicit dependency on the metrics for everything rendering the client, which is also ok.

PR #38, opened 2018-10-01
Add RHD credentials to deploy configs

jfchevrette:
This adds a sample secret resource as well as link to the values as env variables in the telemeter-server deployment.

### DIFF
--- a/deploy/telemeter-server.yaml
+++ b/deploy/telemeter-server.yaml
@@ -96,6 +96,21 @@ objects:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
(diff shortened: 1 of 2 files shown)


s-urbaniak:
No concerns from my side! :+1: 
cc @squat

s-urbaniak:
just to verify: this is in lockstep with https://github.com/app-sre/saas-telemeter ?

jfchevrette:
> just to verify: this is in lockstep with https://github.com/app-sre/saas-telemeter ?

It is. It's also probably worth noting that production configs are deployed out-of-band at the moment.

PR #43, opened 2018-10-05
add jsonnet for prometheus

squat:
This PR adds the jsonnet files necessary to generate the prometheus statefulset that scrapes the telemeter server (AKA the "insights" prometheus). The PR also commits the generated yaml files.

I have tested deploying the generated files against a 3.10 openshift cluster and everything works as expected :tada:!!

In order to finish the work of scraping the telemeter server, we need to create a ServiceMonitor manifest for it.

cc @s-urbaniak @jmelis @pbergene

### DIFF
--- a/jsonnet/telemeter/prometheus/kubernetes/kubernetes.libsonnet
+++ b/jsonnet/telemeter/prometheus/kubernetes/kubernetes.libsonnet
@@ -1,0 +1,332 @@
+local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
+
+{
+  _config+:: {
+    namespace: 'telemeter',
+
+    versions+:: {
+      openshiftOauthProxy: 'v1.1.0',
+      prometheus: 'v2.3.2',
+    },
+
+    imageRepos+:: {
+      openshiftOauthProxy: 'openshift/oauth-proxy',
+      prometheus: 'quay.io/prometheus/prometheus',
+    },
+
+    prometheus+:: {
+      name: 'telemeter',
+      replicas: 2,
+      rules: { groups: [] },
+      htpasswdAuth: '',
+      sessionSecret: '',
+    },
+  },
+
+  prometheus+:: {
+    // The proxy secret is there to encrypt session created by the oauth proxy.
+    proxySecret:
+      local secret = k.core.v1.secret;
+
+      secret.new('prometheus-%s-proxy' % $._config.prometheus.name, {
+        session_secret: std.base64($._config.prometheus.sessionSecret),
+      }) +
+      secret.mixin.metadata.withNamespace($._config.namespace) +
+      secret.mixin.metadata.withLabels({ 'k8s-app': 'prometheus-' + $._config.prometheus.name }),
+
+    htpasswdSecret:
+      local secret = k.core.v1.secret;
+
+      secret.new('prometheus-%s-htpasswd' % $._config.prometheus.name, {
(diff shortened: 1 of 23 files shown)


s-urbaniak:
Looks great! :tada: Leaving the final word to the service delivery team @pbergene

jfchevrette:
Though I don't have a lot of knowledge around jsonnet, the generated yaml looks good. :+1:

squat:
@jfchevrette one question: do you need the target namespace for the Prometheus server to be different? Or can it remain as “openshift-monitoring”?

jfchevrette:
@squat we cannot deploy anything in the kube- or openshift- namespaces. It needs to be in our own namespace(s). Ultimately we are not going to be cluster-admins either so we need to think in terms of what a regular cluster tenant can do.

squat:
Ok good to know, this is why I was asking. I can change the namespace to “telemeter-server” or something then

jsonnet/telemeter/prometheus/kubernetes/kubernetes.libsonnet
```diff
@@ -0,0 +1,339 @@
...
+      prometheus: 'v2.3.2',
+    },
+
+    imageRepos+:: {
+      openshiftOauthProxy: 'openshift/oauth-proxy',
+      prometheus: 'quay.io/prometheus/prometheus',
+    },
+
+    prometheus+:: {
+      namespace: 'telemeter',
```
squat: cc @metalmatze from our discussion of Friday.

All of the prometheus-telemeter resources need to end up in a different namespace than the telemeter-client and other resources generated by jsonnet, thus the specific declaration here. I added an `overrideNamespace` function to `main.jsonnet` that would modify the namespace of rendered resources, but after implementing it I realized that the namespace can be used in more places than just `metadata.namespace`: it is also used in rolebindings and in servicemonitors, so the `overrideNamespace` function was not sufficient and I reverted it.

What are your thoughts?

metalmatze: Probably generating this from a completely new entrypoint is your only real option now. :confused: Sorry.

squat: :tada: done

s-urbaniak:
this LGTM :+1: ping @jfchevrette: as @squat mentioned is the `telemeter` namespace ok for you?

jfchevrette:
@s-urbaniak can we leave the namespace setting out so we can provision this under any namespace we choose rather than hardcode a namespace? I'm thinking we may eventually want to deploy telemeter under a stage or test namespace to validate features and bugfixes, etc...

squat:
@jfchevrette would it be better to compile all the manifests into a `template` manifest where the namespace is templatable (similar to what we have done for telemeter-server)?

manifests/prometheus/list.yaml
```diff
@@ -0,0 +1,261 @@
...
+      interval: 30s
+      port: https
+      scheme: https
+      tlsConfig:
+        caFile: /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
+        serverName: prometheus-telemeter.telemeter.svc
+    selector:
+      matchLabels:
+        prometheus: telemeter
+parameters:
```
squat: @jfchevrette the repo will now contain a Template manifest in `manifests/prometheus/list.yaml` that can be templated as needed and deployed.

PR #46, opened 2018-10-08
add telemeter server jsonnet

squat:
This PR adds jsonnet configuration to generate the telemter server manifests. In addition to the regular manifests, it also generates a list.yaml, which is a composite of all of the telemeter server manifests, as this is the preferred deployment artifact for the app-sre team. This list is a template and includes templatable parameters.

Note: the jsonnet now has two entrypoints, one for the client and one for the server. This is needed because each of these groups of manifests needs to be deployed to a different namespace but only 1 top-level `_config.namespace` is available.

cc @s-urbaniak @metalmatze


jsonnet/telemeter/server/kubernetes/kubernetes.libsonnet
```diff
@@ -0,0 +1,167 @@
...
+          readinessProbe: {
+            httpGet: {
+              path: '/healthz/ready',
+              port: externalPort,
+              scheme: 'HTTPS',
+            },
+          },
+        };
+
+      statefulSet.new('telemeter-server', 3, [telemeterServer], {}, podLabels) +
```
s-urbaniak: just for readability, can we mixin those settings to make those named-parameter-like? for the reader it is very unclear what those parameters mean.

squat: Unfortunately, the ksonnet library does not specify defaults in the `new` constructor functions. That means two that eliminating the constructor leaves us with two bad options where we replace one single constructor line with 6+ additional mixins:
1. replace the constructor with another constructor with empty values and then manually override each field that would have been set. This is more confusing and very verbose: 
    ```
    statefulSet.new('', 0, [], {}, {}) +
    statefulSet.mixin.metadata.withName(name) + 
    statefulSet.mixin.spec.withReplicas(replicas) +
    statefulSet.mixin.spec.template.spec.withContainers(containers) +
    statefulSet.mixin.spec.withVolumeClaimTemplates(volumeClaims) +
    statefulSet.mixin.spec.template.metadata.withLabels(podLabels)
    ```

2. replace the constructor with all of the mixins plus manually specify the API version and kind, since these are not exposed in ksonnet:
    ```
    {apiVersion: "apps/v1beta2"} + 
    {kind: "StatefulSet"} + 
    statefulSet.mixin.metadata.withName(name) + 
    statefulSet.mixin.spec.withReplicas(replicas) +
    statefulSet.mixin.spec.template.spec.withContainers(containers) +
    statefulSet.mixin.spec.withVolumeClaimTemplates(volumeClaims) +
    statefulSet.mixin.spec.template.metadata.withLabels(podLabels)
    ````

Having to manually specify the API version + kind is error prone and seems very one-off; and specifying the empty constructor just to get the correct basic mixin is confusing and harder to understand. For these reasons, I don't think it is a good idea to eschew the constructor. What do you think?

metalmatze: I think it's enough to have something like `$._config.telemeterServer.replicas` and pass that one into the `.new()` function.

s-urbaniak: @squat thanks a lot for the analysis! I was hoping the statefulset would have a non-param constructor like ingress does:

https://github.com/s-urbaniak/k8s-hetzner/blob/f849a24c32b8854bf3cfeaadc971025e6cd9b4a4/jsonnet/telemeter.jsonnet#L8-L14

This is no way is a stopper, it would be just nice to hint somehow the semantic meaning of those parameters.

squat: @s-urbaniak you got it. And to clarify: yes ingress and many of the more basic "building" blocks of K8s (pods, bindings, endpoints, etc) have ksonnet constructors that take 0 parameters, but the higher-level objects like Deployments and StatefulSets are parameterized. jsonnet allows for functions to have default values but this is not leveraged currently in constructors in ksonnet.

PR #47, opened 2018-10-10
telemeter refactorings

s-urbaniak:


s-urbaniak:
@squat PTAL, we can also iterate offline as this PR is huge

pkg/http/roundtripper.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (rt *debugRoundTripper) RoundTrip(req *http.Request) (res *http.Response, err error) {
+	reqd, _ := httputil.DumpRequest(req, false)
+	reqBody := bodyToString(&req.Body)
+
+	res, err = rt.next.RoundTrip(req)
+	if err != nil {
+		log.Println(err)
+		return
```
squat: If a real error is received from the underlying roundtripper, then we should not ignore the it; we should log it and then return it so it is surfaced.

pkg/http/roundtripper.go
```diff
@@ -0,0 +1,81 @@
...
+
+type debugRoundTripper struct {
+	next http.RoundTripper
+}
+
+func NewDebugRoundTripper(next http.RoundTripper) *debugRoundTripper {
+	return &debugRoundTripper{next}
+}
+
+func (rt *debugRoundTripper) RoundTrip(req *http.Request) (res *http.Response, err error) {
```
squat: nit: using named returns is a little out of place in this code base. we should follow up with a cleanup to be consistent

pkg/authorize/roundtripper.go
```diff
@@ -0,0 +1,49 @@
...
+	"net/http"
+	"net/url"
+)
+
+type ServerRotatingRoundTripper struct {
+	endpoint     *url.URL
+	initialToken string
+	tokenStore   tokenStore
+
+	wrapper http.RoundTripper
```
squat: wrapper communicates the opposite relationship between the ServerRotatingRoundTripper and the underlying RoundTripper. Something like `inner` would be more appropriate. In any case, we should decide on something and use it consistently also with the DebugRoundTripper

cmd/telemeter-client/main.go
```diff
@@ -292,9 +312,9 @@ func (o *Options) Run() error {
 
 	if len(o.Listen) > 0 {
 		handlers := http.NewServeMux()
-		telemeterhttp.AddDebug(handlers)
-		telemeterhttp.AddHealth(handlers)
-		telemeterhttp.AddMetrics(handlers)
+		telemeterhttp.DebugRoutes(handlers)
```
squat: as we discussed, we need to follow up and do some cleanup here

cmd/telemeter-server/main.go
```diff
@@ -211,13 +191,21 @@ func (o *Options) Run() error {
 		}
 		authorizeURL = u
 
+		var transport http.RoundTripper
+
+		transport = &http.Transport{
+			Dial:                (&net.Dialer{Timeout: 10 * time.Second}).Dial,
```
squat: The `Dial` field has been deprecated since a few Golang releases ago in favor of `DialContext` [0]. We should clean this up in a follow up as well.

[0] https://golang.org/pkg/net/http/#Transport

cmd/telemeter-server/main.go
```diff
@@ -317,28 +316,38 @@ func (o *Options) Run() error {
...
 	// configure the authenticator and incoming data validator
-	auth := server.New(o.PartitionKey, authorizeURL, authorizeClient, o.TokenExpireSeconds, signer, o.Labels)
-	validator := untrusted.NewValidator(o.PartitionKey, o.Labels, o.LimitBytes, 24*time.Hour)
+	var clusterAuth authorize.ClusterAuthorizer
+	clusterAuth = authorize.ClusterAuthorizerFunc(stub.Authorize)
+	if authorizeURL != nil {
+		clusterAuth = tollbooth.NewAuthorizer(authorizeClient, authorizeURL)
+	}
+
+	auth := jwt.NewAuthorizeClusterHandler(o.PartitionKey, o.TokenExpireSeconds, signer, o.Labels, clusterAuth)
```
squat: in a follow up we really need to slim down this tremendous `Run` method

pkg/store/instrumented/instrumented.go
```diff
@@ -0,0 +1,53 @@
...
+		return s.target.ReadMetrics(ctx, minTimestampMs, fn)
+	}
+	return nil
+}
+
+func (s *instrumented) WriteMetrics(ctx context.Context, partitionKey string, families []*clientmodel.MetricFamily) error {
+	if s.target != nil {
+		err := s.target.WriteMetrics(ctx, partitionKey, families)
+		if err != nil {
+			return err
```
squat: In a follow up we should add another metric here that counts the write errors

pkg/store/instrumented/instrumented.go
```diff
@@ -0,0 +1,53 @@
...
+	}, []string{"phase"})
+)
+
+func init() {
+	prometheus.MustRegister(metricSamples)
+}
+
+type instrumented struct {
+	target store.Store
+	gauge  prometheus.Gauge
```
squat: This is should be a counter, not a gauge. The number of written metrics should never need to be reset (and in fact it is never reset in this code), only incremented monotonically.

pkg/authorize/roundtripper.go
s-urbaniak: agreed, I didn't change the naming in the refactoring, but `next` would be more appropriate.

cmd/telemeter-client/main.go
s-urbaniak: yes :+1:

cmd/telemeter-server/main.go
s-urbaniak: indeed, good catch, this needs a follow-up.

cmd/telemeter-server/main.go
s-urbaniak: yes :+1:

pkg/store/instrumented/instrumented.go
s-urbaniak: sounds good, ack!

pkg/store/instrumented/instrumented.go
s-urbaniak: agreed, this refactoring didn't change the metrics type, but we could do it here. I'll add the change to a counter in this PR.

pkg/store/instrumented/instrumented.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/openshift/telemeter/pkg/store"
+	"github.com/prometheus/client_golang/prometheus"
+	clientmodel "github.com/prometheus/client_model/go"
+)
+
+var (
+	metricSamples = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "telemeter_server_samples",
+		Help: "Tracks the number of samples processed by this server.",
+	}, []string{"phase"})
```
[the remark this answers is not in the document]
s-urbaniak: the existing code tracks the following `phase` values: `received`, `disk`, and `memory`

s-urbaniak:
@squat squashed the commits into one, PTAL.

pkg/authorize/jwt/client_authorizer.go
```diff
@@ -0,0 +1,97 @@
...
+	"crypto"
+	"encoding/base64"
+	"encoding/json"
+	"strings"
+
+	"github.com/openshift/telemeter/pkg/authorize"
+	"gopkg.in/square/go-jose.v2/jwt"
+)
+
+// NewClientAuthorizer authenticates tokens as JWT tokens produced by JWTTokenGenerator
```
squat: `NewClientAuthorizer` does not actually authenticate anything, right? It is only a constructor. We should follow up and fix these docs since they are a little misleading. We could write this for the godoc for the returned type.

Also, since this type is `...Authorizer` we should explicitly document what authorization check is being performed. Is the policy for authorization simply verifying the issuer and signer, and anyone matching this rule is fully authorized for all access? Otherwise we are only authenticating.

s-urbaniak: ack, yes, you have a good point, I shifted some godocs around in order to not loose any language from old code. In the next pass I definitely want to complete godoc :+1:

pkg/authorize/jwt/client_authorizer.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		return nil, false, err
+	}
+
+	return client, true, nil
+}
+
+// hasCorrectIssuer returns true if tokenData is a valid JWT in compact
+// serialization format and the "iss" claim matches the iss field of this token
+// authenticator, and otherwise returns false.
```
squat: in a followup we should ensure we are consistent between terminology. Here we are calling it a `token authenticator` but the method receiver is a pointer to a `clientAuthorizer`. Clearly an authorizer is not the same as an authenticator so we should choose one or the other

pkg/authorize/jwt/client_authorizer.go
```diff
@@ -0,0 +1,97 @@
...
+		return false
+	}
+	claims := struct {
+		// WARNING: this JWT is not verified. Do not trust these claims.
+		Issuer string `json:"iss"`
+	}{}
+	if err := json.Unmarshal(payload, &claims); err != nil {
+		return false
+	}
+	if claims.Issuer != j.iss {
```
squat: as a shortcut, we could return `claims.Isser == j.iss`

s-urbaniak: ack, good catch :+1: let's do this in a follow-up.

pkg/authorize/jwt/signer.go
```diff
@@ -0,0 +1,91 @@
...
+		Claims(claims).
+		Claims(&jwt.Claims{
+			Issuer: j.iss,
+		}).
+		CompactSerialize()
+}
+
+func multipleErrors(errs []error) error {
+	if len(errs) > 1 {
+		return listErr(errs)
```
squat: Nothing in this package actually uses the unexported `listErr` type, or type asserts a returned error into a `listErr`, so we may want to consider eliminating this type and always returning:
```go
var messages []string
for _, err := range errs {
	messages = append(messages, err.Error())
}
errors.New(strings.Join(messages, "; "))
```

Especially given that the `multipleErrors` func is only ever used in one place.

pkg/authorize/jwt/claims.go
```diff
@@ -0,0 +1,32 @@
...
+type telemeter struct {
+	Labels map[string]string `json:"labels,omitempty"`
+}
+
+type privateClaims struct {
+	Telemeter telemeter `json:"telemeter.openshift.io,omitempty"`
+}
+
+func Claims(subject string, labels map[string]string, expirationSeconds int64, audience []string) (*jwt.Claims, interface{}) {
+	now := now()
```
squat: this is not necessary, IMO. Just use `now=time.Now()` here.

pkg/authorize/jwt/validator.go
```diff
@@ -0,0 +1,77 @@
...
+var _ = Validator(&validator{})
+
+func (v *validator) Validate(_ string, public *jwt.Claims, privateObj interface{}) (*authorize.Client, error) {
+	private, ok := privateObj.(*privateClaims)
+	if !ok {
+		log.Printf("jwt validator expected private claim of type *privateClaims but got: %T", privateObj)
+		return nil, errors.New("token could not be validated")
+	}
+	err := public.Validate(jwt.Expected{
+		Time: now(),
```
squat: here too, we should just use `time.Now`

pkg/authorize/jwt/signer.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (errs listErr) Error() string {
+	var messages []string
+	for _, err := range errs {
+		messages = append(messages, err.Error())
+	}
+	return "multiple errors: " + strings.Join(messages, ", ")
+}
+
+func now() time.Time {
```
squat: We should clean this up in a refactor, there is no need for this

pkg/authorize/jwt/validator.go
```diff
@@ -0,0 +1,77 @@
...
+	"gopkg.in/square/go-jose.v2/jwt"
+)
+
+// Validator is called by the JWT token authentictaor to apply domain specific
+// validation to a token and extract user information.
+type Validator interface {
+	// Validate validates a token and returns user information or an error.
+	// Validator can assume that the issuer and signature of a token are already
+	// verified when this function is called.
+	Validate(tokenData string, public *jwt.Claims, private interface{}) (*authorize.Client, error)
```
squat: the GoDoc for the `Validator` interface says this called by the token authenticator, but the type returned by this method is a `*authorize.Client`, maybe we should refactor this into a more generic `auth` package, or split authn and authz into two distinct packaged

pkg/authorize/tollbooth/mock.go
```diff
@@ -81,3 +83,9 @@ func Write(w http.ResponseWriter, statusCode int, resp interface{}) error {
 	w.Write(data)
 	return nil
 }
+
+func fnvHash(text string) string {
```
squat: we have two identical implementations of this, one in authorize/stub and another in authorize/tollbooth.mock. Maybe we should mark one as exported and share it.

pkg/authorize/tollbooth/tollbooth.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil, err
+	}
+	response := &clusterRegistration{}
+	if err := json.Unmarshal(body, response); err != nil {
+		return nil, err
+	}
+	return response, nil
+}
+
+type errWithCode struct {
```
squat: Nothing  ever type asserts an error to this type, we should eliminate this in a follow up

pkg/cluster/cluster.go
```diff
@@ -208,37 +190,56 @@ func (c *DynamicCluster) getNodeForKey(partitionKey string) (string, bool) {
...
+//
+// See github.com/hashicorp/memberlist#EventDelegate.NotifyLeave
 func (c *DynamicCluster) NotifyLeave(node *memberlist.Node) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 	log.Printf("[%s] node left %s", c.name, node.Name)
 	c.ring.RemoveNode(node.Name)
 }
+
+// NotifyUpdate is the callback that is invoked when a node to have updated.
```
squat: there seems to be a type in this new doc here. Should be `node is detected to have been updated`

pkg/cluster/memberlist.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/hashicorp/memberlist"
+)
+
+type delegate interface {
+	memberlist.EventDelegate
+	memberlist.Delegate
+}
+
+func NewMemberlist(name, addr string, secret []byte, verbose bool, d delegate) (*memberlist.Memberlist, error) {
+	if len(secret) != 32 {
```
squat: in a follow up we need to put magic numbers into consts

pkg/cluster/memberlist.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	host, portString, err := net.SplitHostPort(addr)
+	if err != nil {
+		return nil, fmt.Errorf("address must be a host:port: %v", err)
+	}
+
+	port, err := strconv.Atoi(portString)
+	if err != nil {
+		return nil, fmt.Errorf("address must be a host:port: %v", err)
```
squat: copy-pasta: the error message here should not be about `host:port` but about `port` not being an integer.

pkg/http/instrument.go
```diff
@@ -0,0 +1,50 @@
...
+	requestsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "http_requests_total",
+			Help: "Tracks the number of HTTP requests.",
+		}, []string{"code", "handler", "method"},
+	)
+)
+
+func init() {
+	prometheus.MustRegister(requestDuration, requestSize, requestsTotal)
```
squat: In a follow up we should move this out of an `init` func, in general `init`s should be avoided IMO, since they don't always play nice. Also, we should avoid using the default prometheus registry in favor of an explicitly created one.

pkg/http/roundtripper.go
```diff
@@ -0,0 +1,81 @@
...
+	reqd, _ := httputil.DumpRequest(req, false)
```
squat: given that this is an extra verbose setting, I think it would be preferable not do discard errors but rather to log them, as they could be useful in debugging errors

pkg/http/roundtripper.go
```diff
@@ -0,0 +1,81 @@
...
+
+func bodyToString(body *io.ReadCloser) string {
+	if *body == nil {
+		return "<nil>"
+	}
+
+	var b bytes.Buffer
+	_, err := b.ReadFrom(*body)
+	if err != nil {
+		panic(err)
```
squat: I don't think there is any reason to panic here. Nothing is forcing the `bodyToString` signature to be `string` rather than `(string, error)`. Especially given that this is _only_ used in the `debugRoundTripper` type, we should return the errors and log them in the calling func.

pkg/validator/validator.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+type Validator struct {
+	partitionKey string
+	labels       map[string]string
+	limitBytes   int64
+	maxAge       time.Duration
+}
+
+// New handles prometheus metrics from end clients that must be assumed to be hostile.
```
squat: This doc has the same issue as the one for the authorizer mentioned above. This constructor func does not `handle prometheus metrics`, as written, it only returns a `*Validator` which can do that handling. More formally, it is only the `ValidateUpload` method that handles prometheus metrics from clients.

pkg/validator/validator.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+}
+
+func (v *Validator) ValidateUpload(ctx context.Context, req *http.Request) (string, metricfamily.Transformer, error) {
+	client, ok := authorize.FromContext(ctx)
+	if !ok {
+		return "", nil, fmt.Errorf("unable to find user info")
+	}
+	if len(client.Labels[v.partitionKey]) == 0 {
+		return "", nil, fmt.Errorf("user data must contain a '%s' label", v.partitionKey)
```
squat: rather than `'%s'`, we should use `%q`

squat:
I know this was a big effort, good work, @s-urbaniak! There are no major issues with this first re-factor attempt, so let's merge and do some serious follow ups shortly after.

s-urbaniak:
@squat agreed, let's switch gears and do factoring follow-ups :+1:

pkg/authorize/tollbooth/tollbooth.go
s-urbaniak: It is used here:
https://github.com/openshift/telemeter/blob/4ca2fbb7298800d1fb37e3507928755210b84afd/pkg/authorize/jwt/handler.go#L73-L86

I am not super happy with the handling though, but we could factor this more nicely in a follow-up.

pkg/cluster/memberlist.go
s-urbaniak: :crystal_ball: :+1:

pkg/cluster/memberlist.go
s-urbaniak: :spaghetti: :+1:

pkg/http/instrument.go
s-urbaniak: yes, that is a very good point, maybe even in the very next follow-up.

pkg/authorize/tollbooth/mock.go
s-urbaniak: addressed in https://github.com/openshift/telemeter/pull/56

PR #48, opened 2018-10-15
deploy: remove deploy directory

squat:
This commit removes the obsolete `deploy` directory. We can delete it
now that it is no longer needed by the ci-operator:
openshift/release#2214.

cc @metalmatze @kakkoyun


s-urbaniak:
looks super! :ok_hand: small question though, as soon as we merge this here, api.ci will fail due to some references to `deploy/telemeter-client-openshift.yaml` as per https://github.com/openshift/release/blob/9c980ba/ci-operator/templates/cluster-launch-e2e.yaml#L132.

I guess we should make this (more or less) in lockstep with a corresponding PR in https://github.com/openshift/release.

metalmatze:
Do we still want this @squat?

squat:
We need to do a quick dig to see if it is still being used in CI as mentioned in https://github.com/openshift/telemeter/pull/48#issuecomment-429855868. Could you TAL?

metalmatze:
I can't find anything related to Telemeter in this file. https://github.com/openshift/release/blob/1ae6f6dc7ad243f1b860b133d6a122e14c269d46/ci-operator/templates/openshift/installer/cluster-launch-installer-e2e.yaml

squat:
This can now safely be removed as the ci-operator no longer depends on it. @metalmatze

squat:
cc @kakkoyun

PR #51, opened 2018-10-16
Create LICENSE

s-urbaniak:
cc @brancz @smarterclayton

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
squat: Do we want to fill this out?

s-urbaniak: haha, my templating skills are really bad, but I see https://github.com/openshift/cluster-monitoring-operator/blob/master/LICENSE#L189 has the same :grin:

squat: ok sounds fine then. I have also seen lots of licenses un-templated, though strictly speaking I don't know how correct that is and what the implications are on license enforcement.

PR #55, opened 2018-10-22
linting: fix issues in telemeter, add linting to CI

s-urbaniak:


s-urbaniak:
@squat PTAL, this is ready for a pass :-)

Makefile
```diff
@@ -17,7 +19,12 @@ build:
 image:
 	imagebuilder -t openshift/telemeter:latest .
 
-check:
+lint: $(GOLANGCI_LINT_BIN)
+	# megacheck fails to respect build flags, causing compilation failure during linting.
+	# instead, use the unused, gosimple, and staticcheck linters directly
+	$(BIN)/golangci-lint run -e '.*Write' -D megacheck -E unused,gosimple,staticcheck
```
s-urbaniak: honestly I am not sure about ignoring `.*Write` method signatures not checking errors. We can enable this, the code just becomes a bit chatty in some places with `_ = Write...`.

pkg/authorize/tollbooth/tollbooth.go
```diff
@@ -59,7 +59,7 @@ func (a *authorizer) AuthorizeCluster(token, cluster string) (string, error) {
 	defer func() {
 		// read the body to keep the upstream connection open
 		if resp.Body != nil {
-			io.Copy(ioutil.Discard, resp.Body)
+			_, _ = io.Copy(ioutil.Discard, resp.Body)
```
squat: We can ignore the byte count but we shouldn’t suppress the error. If there is one, let’s at the very least log it

pkg/metricsclient/metricsclient.go
```diff
@@ -118,13 +118,13 @@ func (c *Client) Send(ctx context.Context, req *http.Request, families []*client
...
 
-	ctx, cancel := context.WithTimeout(context.Background(), c.timeout)
+	ctx, cancel := context.WithTimeout(ctx, c.timeout)
 	req = req.WithContext(ctx)
 	defer cancel()
 
 	return withCancel(ctx, c.client, req, func(resp *http.Response) error {
 		defer func() {
-			io.Copy(ioutil.Discard, resp.Body)
+			_, _ = io.Copy(ioutil.Discard, resp.Body)
```
squat: Same here

Makefile
squat: I don’t much like the idea of ignoring the errors in `.Write...`. We can ignore the byte counts but even if we don’t return the error, at least log it. Go2’s error handling will make this feel less “chatty” 😉 the insight into write failures can be very informative about network, memory, or disk issues.

s-urbaniak:
@squat addressed comments, PTAL

pkg/fnv/hash.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"fmt"
+	"hash"
+	"hash/fnv"
+	"strconv"
+)
+
+// Hash hashes the given text using a 64-bit FNV-1a hash.Hash.
+func Hash(text string) (string, error) {
```
squat: These hash changes are related to the comment [0] in the refactoring PR but I think they are out of scope for this linting PR. Let's move this into it's separate pull request so we can keep the PR as one cohesive unit of work centered around linting.

[0] https://github.com/openshift/telemeter/pull/47#discussion_r225924456

s-urbaniak: oki, agreed

s-urbaniak:
@squat removed the commit, PTAL

PR #56, opened 2018-10-23
pkg/authorize: use centralized fnv hash

s-urbaniak:
Fixes https://github.com/openshift/telemeter/pull/47#discussion_r225924456

/cc @squat


s-urbaniak:
@squat no worries! please re-submit the /lgtm as it seems to be misspelled.

PR #57, opened 2018-10-24
Default to url

squat:
Add a default URL value for the `--to` flag of the telemeter client.

cc @s-urbaniak

### DIFF
--- a/cmd/telemeter-client/main.go
+++ b/cmd/telemeter-client/main.go
@@ -192,7 +192,7 @@ func (o *Options) Run() error {
 	if len(o.ToUpload) > 0 {
 		to, err = url.Parse(o.ToUpload)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/telemeter-client/main.go
```diff
@@ -192,7 +192,7 @@ func (o *Options) Run() error {
 	if len(o.ToUpload) > 0 {
 		to, err = url.Parse(o.ToUpload)
 		if err != nil {
-			return fmt.Errorf("--to is not a valid URL: %v", err)
+			return fmt.Errorf("--to-upload is not a valid URL: %v", err)
```
s-urbaniak: the cli flag is still just `--to`, no?

s-urbaniak: ah, that makes sense, thanks for fixing! :+1:

squat: :+1:

PR #58, opened 2018-10-24
jsonnet,metrics: move metrics to independent json

squat:
Move the labels to match metrics to an independent JSON file. This makes
it easier for contributors to inspect and modify the labels used to
select metrics without needing to understand jsonnet. The repository
root contains a symlink to the labels file for ease of discovery.

cc @s-urbaniak @brancz

### DIFF
--- a/jsonnet/telemeter/client/metrics.json
+++ b/jsonnet/telemeter/client/metrics.json
@@ -1,0 +1,8 @@
+[
+    "{__name__=\"up\"}",
+    "{__name__=\"openshift_build_info\"}",
+    "{__name__=\"machine_cpu_cores\"}",
+    "{__name__=\"machine_memory_bytes\"}",
+    "{__name__=\"etcd_object_counts\"}",
+    "{__name__=\"ALERTS\",alertstate=\"firing\"}"
+]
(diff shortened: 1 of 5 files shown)


jsonnet/telemeter/client/metrics.json
```diff
@@ -0,0 +1,8 @@
+[
+    "{__name__=\"up\"}",
```
brancz: It would be valid json if we used single quotes around these? If so maybe a readability addition.

Usually wouldn't be this nit-picky but given that this is probably going to be a point of collaboration with other teams, this is probably something we want to make as easy as possible.

s-urbaniak: yes, this is a good point actually.

squat: JSON requires double quotes around all string values :/. Or do you mean single quotes around the label values?

brancz: Sorry yes, should have been more explicit. Label selectors typically allow using single quotes. But now that you mention it, I actually have no idea about the federation endpoint.

squat: I'll run a quick experiment to check :+1:

PR #63, opened 2018-10-26
jsonnet/telemeter/client: add support for salt

squat:
This commit adds support for using a salt to anonymize metrics reported
by the client.

cc @s-urbaniak


jsonnet/telemeter/client/kubernetes/kubernetes.libsonnet
```diff
@@ -134,7 +137,9 @@ local securePort = 8443;
 
       secret.new(secretName, {
         [matchFileName]: std.base64(std.join('\n', $._config.telemeterClient.matchRules)),
+        salt: std.base64($._config.telemeterClient.salt),
         to: std.base64($._config.telemeterClient.to),
+        token: std.base64($._config.telemeterClient.token),
```
squat: By default these will be empty but they are added here for correctness and completeness.

PR #64, opened 2018-10-26
ttl based memory store

s-urbaniak:
This PR:
1. simplifies the store.Store interface (as discussed)
2. adds ttl capabilities to the memory store
3. adapts the telemeter server deployment accordingly

Fixes MON-346


cmd/telemeter-server/main.go
```diff
@@ -100,7 +99,7 @@ func main() {
 	cmd.Flags().StringVar(&opt.TLSKeyPath, "tls-key", opt.TLSKeyPath, "Path to a private key to serve TLS for external traffic.")
 	cmd.Flags().StringVar(&opt.TLSCertificatePath, "tls-crt", opt.TLSCertificatePath, "Path to a certificate to serve TLS for external traffic.")
 
-	cmd.Flags().StringVar(&opt.StorageDir, "storage-dir", opt.StorageDir, "The directory to persist incoming metrics. If not specified metrics will only live in memory.")
+	cmd.Flags().StringVar(&opt.StorageDir, "storage-dir", opt.StorageDir, "DEPRECATED: The directory to persist incoming metrics. If not specified metrics will only live in memory.")
```
squat: I don't think we need to worry about breaking people. No one is using telemeter yet. This option has no utility in the code any more, so let's simply remove it.

pkg/cluster/cluster.go
```diff
@@ -290,7 +289,10 @@ func (c *DynamicCluster) handleMessage(data []byte) error {
 		if len(families) == 0 {
 			return nil
 		}
-		return c.store.WriteMetrics(c.ctx, header.PartitionKey, families)
+		return c.store.WriteMetrics(c.ctx, &store.PartitionedMetrics{
```
squat: these changes were nice :)

pkg/http/server/server_test.go
```diff
@@ -32,7 +32,10 @@ func family(name string, timestamps ...int64) *clientmodel.MetricFamily {
 func storeWithData(data map[string][]*clientmodel.MetricFamily) store.Store {
 	s := memstore.New()
 	for k, v := range data {
-		if err := s.WriteMetrics(context.TODO(), k, v); err != nil {
+		if err := s.WriteMetrics(context.TODO(), &store.PartitionedMetrics{
```
squat: nit, the `if err := ...; err != nil` pattern is nice for one-liners, but for multi-line statements becomes quite hard to read. I would suggest breaking this into multiple lines

pkg/store/diskstore/diskstore.go
```diff
@@ -34,8 +35,10 @@ func lastFile(files []os.FileInfo) os.FileInfo {
 	return nil
 }
 
-func (s *diskStore) ReadMetrics(ctx context.Context, minTimestampMs int64, fn func(partitionKey string, families []*clientmodel.MetricFamily) error) error {
-	return filepath.Walk(s.path, func(path string, info os.FileInfo, err error) error {
+func (s *diskStore) ReadMetrics(ctx context.Context, minTimestampMs int64) ([]*store.PartitionedMetrics, error) {
```
squat: Telemeter does not use this code the diskstore at all and has no entry-point to it, so I would suggest we simply remove this package entirely.

pkg/store/memstore/memstore.go
```diff
@@ -22,30 +28,84 @@ func New() *memoryStore {
...
+				s.cleanup(time.Now)
+			case <-ctx.Done():
+				ticker.Stop()
+				return
+			}
+		}
+	}()
+}
+
+func (s *memoryStore) cleanup(now func() time.Time) {
```
squat: why not simply pass a `time.Time`? Or better yet, eliminate the parameter all together? I think this implementation has two issues:
1. Likely to cause confusion because `now` is actually the time when the function is called in the for-loop, not when the `cleanup` func is called
2. It is cumbersome for calling cleanup with a time other than now, as we have to do something like
    ```go
    s.cleanup(func() time.Time {return time.Now().Add(-5*time.Minute)})
    ```
    Instead of something like
    ```go
    s.cleanup(time.Now().Add(-5*time.Minute))
    ```

Additionally, the semantics of this memory-store are clear: it is TTL-based. This has little meaning for TTLs relative to times other than now. In any case, the TTL could be a configurable parameter.

To simplify here, I propose eliminating the `now` parameter entirely.

pkg/store/memstore/memstore.go
```diff
@@ -2,17 +2,23 @@ package memstore
...
 	"context"
+	"math"
 	"sync"
+	"time"
 
+	"github.com/openshift/telemeter/pkg/store"
 	clientmodel "github.com/prometheus/client_model/go"
 )
 
+const ttl = 20 * time.Minute
```
squat: One this strikes me as funny, in the telemeter server, we hardcode the maximum sample age of a federated metric to be:
```go
// maxSampleAge is the maximum age of a sample that we can report via federation.
const maxSampleAge = 10 * time.Minute
```

This means, we never report any metrics older than 10 minutes, but with this TTL we will still hold on to metrics as old as 20 minutes. Why? Why do we want the ability to hold on to metrics that we will never report? Furthermore, the semantics of the opposite, that is, having a TTL that is shorter than the max age are likewise odd, e.g. we could have a TTL of 5 minutes and a max age of 10 minutes.

Why not use the max age as the TTL value for cleanups?

squat:
Looks good overall, @s-urbaniak! I have some comments and questions about semantics and implementation. Sorry they are so long. Please take a look :)

cmd/telemeter-server/main.go
s-urbaniak: sounds good :+1:

pkg/http/server/server_test.go
s-urbaniak: will do

pkg/store/diskstore/diskstore.go
s-urbaniak: deleted code is the best code :+1:

pkg/store/memstore/memstore.go
s-urbaniak: I use `now` only for testing only to pass a different "now" function, there is no other meaning.

The invoking function in non-tests is as simple as:
https://github.com/openshift/telemeter/blob/1dfcc6840e22399707cf64d60111426661a7efcf/pkg/store/memstore/memstore.go#L41.

But I am totally fine with passing `time.Time()` instead.

For testability I would still recommend injecting it so we don't have potential flakes in tests.

pkg/store/memstore/memstore.go
s-urbaniak: I was actually thinking the same. The only reason to keep it here was to have some additional margin for the data to live in the store in the case the consumer has to retry moving `maxSampleAge` further out in time. I am actually fine with merging it, wdyt?

pkg/store/memstore/memstore.go
squat: Ok, then let's do that. Let's make this accept a `time.Time` instead.

pkg/store/memstore/memstore.go
squat: Even if the consumer had to retry fetching metrics, if the time of the metric was greater than the max age, then it doesn't matter how long the TTL is, the metric will not be returned. If the metric timestamp is less than the max age then the metrics will be returned. So in this case there is still no benefit to having a TTL that is longer than the max age.

I think we should merge this for now to simplify; there are a lot of weird edge cases and unspecified behaviors currently depending on the relative values of these constants.

s-urbaniak:
@squat addressed comments, PTAL

pkg/store/memstore/memstore.go
s-urbaniak: agreed, merged :+1:

squat:
:tada:

PR #67, opened 2018-10-29
Adds build-deploy.sh script for CI integration

jmelis:
This script using single stage build and pushes to quay.io

### DIFF
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -5,9 +5,9 @@ set -exv
 IMAGE_BUILD="telemeter-build"
 DOCKERFILE_BUILD="dockerfiles/Dockerfile.build"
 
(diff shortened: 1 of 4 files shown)


build_deploy.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -exv
+
+IMAGE_BUILD="telemeter-build"
+DOCKERFILE_BUILD="dockerfiles/Dockerfile.build"
+
+IMAGE_DEPLOY="quay.io/app-sre/app-interface"
```
squat: This variable is never used.

squat: Is the image name `app-interface` desired? or should it be `telemeter`?

PR #72, opened 2018-10-30
 jsonnet: do not render volumeClaimTemplates

### LINKED ISSUE #71: Remove volumeClaimTemplates from StatefulSet manifest
Any attempts to `oc apply` an existing StatefulSet will fail if the new template
contains `volumeClaimTemplates` with the following message:

```
The StatefulSet "telemeter-server" is invalid: spec: Forbidden: updates
to statefulset spec for fields other than 'replicas', 'template', and
'updateStrategy' are forbidden.
```

We believe the underlying problem to be a bug in k8s. In this case, the
generated StatefulSet template contains `volumeClaimTemplates = []`. When the
StatefulSet is created from scratch, that value is removed from the
representation as it's an empty array. However, upon running `oc apply` again
with the same template the validation fails as it believes we're trying to
change the `volumeClaimTemplates` field.

Would it be possible for you to tweak the jsonnet generation so this value is
removed?

s-urbaniak:
cc @jmelis @squat 

Fixes #71

### DIFF
--- a/jsonnet/telemeter/server/kubernetes/kubernetes.libsonnet
+++ b/jsonnet/telemeter/server/kubernetes/kubernetes.libsonnet
@@ -99,7 +99,12 @@ local clusterPort = 8081;
       statefulSet.mixin.spec.withPodManagementPolicy('Parallel') +
       statefulSet.mixin.spec.withServiceName('telemeter-server') +
       statefulSet.mixin.spec.template.spec.withServiceAccountName('telemeter-server') +
-      statefulSet.mixin.spec.template.spec.withVolumes([localVolume, tlsVolume]),
+      statefulSet.mixin.spec.template.spec.withVolumes([localVolume, tlsVolume]) +
+      {
+        spec+: {
+          volumeClaimTemplates:: null,
+        },
+      },
 
     secret:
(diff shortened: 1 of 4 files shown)


jsonnet/telemeter/server/kubernetes/kubernetes.libsonnet
```diff
@@ -99,7 +99,12 @@ local clusterPort = 8081;
       statefulSet.mixin.spec.withPodManagementPolicy('Parallel') +
       statefulSet.mixin.spec.withServiceName('telemeter-server') +
       statefulSet.mixin.spec.template.spec.withServiceAccountName('telemeter-server') +
-      statefulSet.mixin.spec.template.spec.withVolumes([localVolume, tlsVolume]),
+      statefulSet.mixin.spec.template.spec.withVolumes([localVolume, tlsVolume]) +
+      {
+        spec +: {
```
squat: could you run `jsonnet fmt` on this file? there should be no space here. I'll make sure that our CI catches jsonnet format error

s-urbaniak: @squat good idea! :+1: PTAL

PR #73, opened 2018-10-30
check jsonnet format

squat:
This PR adds two new targets to the Makefile: `fmt`, and `check-fmt`. These format the jsonnet and test the jsonnet code format, respectively. `check-fmt` will now be required in CI.


s-urbaniak:
@squat :-(
```
ERROR: the Go implementation currently does not support jsonnet fmt
```

squat:
Ah, true :( so sad

PR #74, opened 2018-11-01
jsonnet/prometheus: configure permissions

s-urbaniak:
This enables permissions (subject access review as well as delegate URLs)
to be configured using saas hurder template variables.

cc @squat

### DIFF
--- a/jsonnet/jsonnetfile.lock.json
+++ b/jsonnet/jsonnetfile.lock.json
@@ -8,7 +8,7 @@
                     "subdir": "jsonnet/telemeter/client"
                 }
             },
-            "version": "e0de8764eb42445e524fd1dd012ecaa7f8e9047d"
+            "version": "5e23aa9d6e369f3ecaf94b728068cf965dabec60"
         },
         {
             "name": "ksonnet",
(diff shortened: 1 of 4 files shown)


jsonnet/telemeter/prometheus/kubernetes/kubernetes.libsonnet
```diff
@@ -20,6 +20,8 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
       rules: { groups: [] },
       htpasswdAuth: '',
       sessionSecret: '',
+      sar: '{"resource": "namespaces", "verb": "get", "resourceName": "telemeter-production", "namespace": "telemeter-production"}',
```
squat: this seems like a bit too much configurability here. I think we can optimize for our actual expected use case :), since we know how it is intended to be deployed. Anyone wild enough to use and want more configurability power this can import the jsonnet and use additional mixins to modify the generated manifests

squat: also, rather than having to cofigure both `$._config.namespace` as well as the SAR so that the namespaces match, the SAR should inherit the namespace to which the object is deployed.

s-urbaniak: ack, see https://github.com/openshift/telemeter/pull/74#discussion_r230011249

s-urbaniak: ah, that's indeed a good idea

s-urbaniak:
@brancz @squat addressed comments, PTAL

brancz:
Looks good. Giving @squat the last call.

squat:
much better

PR #76, opened 2018-11-02
cmd/client,pkg: add reload functionality

squat:
This commit restructures the main.go file of the Telemeter client to
move all of the configuration logic for the worker into the `forwarder`
package. This gives us the ability to reload the worker with new
configuration whenever we like.

This commit also adds a reload HTTP endpoint and the ability to reload
configuration on SIGHUP.

A follow up will add the config reloader container to the deployment.

cc @s-urbaniak


pkg/forwarder/forwarder.go
```diff
@@ -40,85 +48,248 @@ func init() {
...
+	interval    time.Duration
 	transformer metricfamily.Transformer
 	rules       []string
 
-	lock        sync.Mutex
-	lastMetrics []*clientmodel.MetricFamily
+	lastMetricsLock sync.Mutex
+	lastMetrics     []*clientmodel.MetricFamily
+	rc              chan struct{}
+	reconfigureLock sync.Mutex
```
s-urbaniak: suggestion: to keep things simpler, I suggest to reduce the amount of concurrency primitives for reloading to just one, namely the `rc` channel (which I guess is the "reconfigure" channel?).

Instead locking `Worker`'s internal state via `reconfigureLock` _and_ signalling the reload via `rc`, I suggest to remove the lock and communicate the need for reconfiguration via the `rc` channel only.

It could be of type `rc chan Config` and the desired new configuration would be received in `Run`'s select https://github.com/openshift/telemeter/pull/76/commits/41fd79ea65bc053054bb5cb1e660daed1117e653#diff-5dc0eebe45084d4d6def4a4bb1cbb0ddR285 block.

Then, instead of mutating `Worker`'s state in the calling go-routine, it would be mutated in the goroutine that executes the for loop in the `Run` method which would prevent the need for additional locking.

The `Reconfigure` method would be as simple as
```
func (w *Worker) Reconfigure(cfg Config) error {
	// Signal a restart to Run func.
	// This must be done in a goroutine to avoid potential deadlock.
	go func() { w.rc <- cfg }()
	return nil
}
```

pkg/forwarder/forwarder.go
```diff
@@ -40,85 +48,248 @@ func init() {
...
+		i++
+	}
+	w.rules = rules
+
+	return &w, nil
+}
+
+// Reconfigure temporarily stops a worker and reconfigures is with the provided Config.
+// Can be run concurrently with `LastMetrics`
+func (w *Worker) Reconfigure(cfg Config) error {
```
s-urbaniak: Do you mind to add a unit test for the `Reconfigure` logic?

cmd/telemeter-client/main.go
```diff
@@ -301,30 +207,115 @@ func (o *Options) Run() error {
...
+
+	worker, err := forwarder.New(cfg)
+	if err != nil {
+		return fmt.Errorf("failed to configure Telemeter client: %v", err)
+	}
 
 	log.Printf("Starting telemeter-client reading from %s and sending to %s (listen=%s)", o.From, o.To, o.Listen)
 
-	go worker.Run()
+	var g run.Group
```
s-urbaniak: :+1: for using `oklog/run` here ! :-)

cmd/telemeter-client/main.go
```diff
@@ -301,30 +207,115 @@ func (o *Options) Run() error {
...
+			worker.Run(ctx)
+			return nil
+		}, func(error) {
+			cancel()
+		})
+	}
+
+	{
+		// Notify and reload on SIGHUP.
+		hup := make(chan os.Signal)
```
s-urbaniak: nit, according to https://godoc.org/os/signal#Notify:
> Package signal will not block sending to c: the caller must ensure that c has sufficient buffer space to keep up with the expected signal rate. For a channel used for notification of just one signal value, a buffer of size 1 is sufficient.

Hence, I suggest to `hup := make(chan os.Signal, 1)`

pkg/forwarder/forwarder.go
squat: That is indeed a nice alternative. However, I see some issues with this solution is that make it not equivalent to the one in the commit as the reconfiguration would no longer be bounded by the Reconfigure func. Effectively, the reconfigure function would now be asynchronous but with the drawback of not notifying the result via a callback or promise. This means that calls to the /-/reconfigure endpoint will prematurely return a 200 before the reconfiguration has actually been applied. Also, it means that errors in a newly loaded configuration cannot be caught and exposed by the reconfigure function, and so they cannot be bubbled up via the HTTP response in a 4xx status code. Finally, it means that unit testing the reconfigure func is not really possible since there are no guarantees on the worker’s state after the function returns. This is why an additional synchronization point is required. We could achieve this in your suggestion by either: adding an additional channel to send back errors due to applying the new config or by changing the chan to accept a tuple of config + callback to notify a caller of the result of applying the new configuration. Neither of these options seems particularly preferable to the current solution. What do you think?

squat: It is also possible to collapse both locks into one single mutex to guard the entire state of the worker

cmd/telemeter-client/main.go
squat: Great suggestion!

pkg/forwarder/forwarder.go
s-urbaniak: agreed, yes, a synchronous feedback of a (un)successful reconfiguration would be lost with my suggestion. But as this could also happen out of band we could still send a new instance of `Worker` via the channel.

I'll leave the final decision to you, but we should take care to lower the amount of concurrency primitives to a minimum.

I'm :+1: on collapsing at least `reconfigureLock` and `lastMetricsLock` into one.

squat: yes, sending the worker struct on the chan was my other thought, I thought it was a little bit hacky, but it would absolutely work :+1:

pkg/forwarder/forwarder.go
```diff
@@ -40,91 +48,249 @@ func init() {
...
-func (w *Worker) LastMetrics() []*clientmodel.MetricFamily {
+// Reconfigure temporarily stops a worker and reconfigures is with the provided Config.
+// Is thread safe and can run concurrently with `LastMetrics` and `Run`.
+func (w *Worker) Reconfigure(cfg Config) error {
+	worker, err := New(cfg)
+	if err != nil {
+		return fmt.Errorf("failed to reconfigure: %v", err)
+	}
+
 	w.lock.Lock()
```
squat: I decided against using the channel to pass configuration to the worker. Instead, I chose to lock here, keep all the reconfiguration inside the `Reconfigure` func, and signal the `Run` func. Otherwise, if we instead passed a Config or Worker down the channel to the `Run` func, the `Reconfigure` func would not fully encapsulate a reconfiguration, meaning that any process that reconfigured a worker and then made a request against it, would be racing against the `Run` func. This keeps things deterministic and testable.

pkg/forwarder/forwarder.go
squat: I decided against using the channel to pass configuration to the worker. Instead, I chose to lock here, keep all the reconfiguration inside the Reconfigure func, and signal the Run func. Otherwise, if we instead passed a Config or Worker down the channel to the Run func, the Reconfigure func would not fully encapsulate a reconfiguration, meaning that any process that reconfigured a worker and then made a request against it, would be racing against the Run func. This keeps things deterministic and testable.

pkg/http/routes.go
```diff
@@ -30,3 +30,27 @@ func MetricRoutes(mux *http.ServeMux) *http.ServeMux {
 	mux.Handle("/metrics", promhttp.Handler())
 	return mux
 }
+
+// ReloadRoutes adds the reload endpoint to a mux.
+func ReloadRoutes(mux *http.ServeMux) (*http.ServeMux, <-chan func(error)) {
```
s-urbaniak: Let's simplify the ReloadRoutes func to `func ReloadRoutes(mux *http.ServeMux, reloader func() error) (*http.ServeMux)` and implement the channel mangling out of band, i.e. in `main`:

```
reloader := func() error {
  errChan := make(chan error)
  reloadChan <- func(err error) {
    errChan <- err
  }
  return <-errChan
}
```

and then the ReloadRoutes func body becomes as simple as:
```
mux.HandleFunc("/-/reload", func(w http.ResponseWriter, req *http.Request) {
		if req.Method != http.MethodPost {
			w.WriteHeader(http.StatusMethodNotAllowed)
			return
		}
		err := reloader()
		if err != nil {
			w.WriteHeader(http.StatusInternalServerError)
			return
		}
	})
```

what do you think?

squat: Sounds great to me :) . This is very nice. It shuffles the complexity around, but keeps the handler quite clean.

squat: You know what, I don't think we actually really need to put the reload handler logic in a runGroup. If we pass the `Reconfigure` func directly to the handler, we can simplify things even more.

s-urbaniak: @squat even better! 👍 the less concurrency, the better!

PR #78, opened 2018-11-06
pkg/store: implement rate limited store

s-urbaniak:
Fixes MON-363

/cc @squat @brancz


brancz:
I like the simplicity of this, and the middleware approach generally looks good as well, but the PR is not clear on what a client will see in terms of an HTTP response, can you elaborate on that? Right now it seems a user would get a 500 error or something along the lines of that (this is a wild guess, as I'm guessing the rate limited error is just treated as an "unhandled" error; please clarify if this is wrong :slightly_smiling_face: ).

s-urbaniak:
@brancz doh, you are totally right, currently the user would receive a 500 right now, I can add a 429 http status code in the http server implementation.

brancz:
That sounds perfect. I'm not a fan of having an error type check on the http handler, but I think that's far better than leaking http errors into the store. I think of it as a "view" to the user so converting the "internal" error to one that is comprehensible to the outside seems like a valid thing to do.

PR #87, opened 2018-11-09
pkg/authorize/jwt: fix handling of required labels

squat:
This commit fixes the handling of required labels in the JWT package.
Formerly, the JWT authorizer constructor accepted a labels parameter,
but this map was never actually used by the type. This commit fixes the
handling of the parameter so that the labels are added to the JWT. This
means that the labels will now be respected as required labels in the
validator.

This commit also cleans up some general label handling. For example, the
validator was had the responsibility of adding additional labels
to the received metrics, when this really has nothing to do with
validation; it is instead a general purpose transformation. This has
been extracted and put instead into the `Post` handler so that the label
addition occurs after successful validation.

cc @s-urbaniak

### DIFF
--- a/cmd/telemeter-server/main.go
+++ b/cmd/telemeter-server/main.go
@@ -35,12 +35,13 @@ import (
 	"github.com/openshift/telemeter/pkg/cluster"
 	telemeter_http "github.com/openshift/telemeter/pkg/http"
 	httpserver "github.com/openshift/telemeter/pkg/http/server"
+	"github.com/openshift/telemeter/pkg/metricfamily"
 	telemeter_oauth2 "github.com/openshift/telemeter/pkg/oauth2"
 	"github.com/openshift/telemeter/pkg/store"
 	"github.com/openshift/telemeter/pkg/store/instrumented"
 	"github.com/openshift/telemeter/pkg/store/memstore"
 	"github.com/openshift/telemeter/pkg/store/ratelimited"
-	"github.com/openshift/telemeter/pkg/validator"
+	"github.com/openshift/telemeter/pkg/validate"
 )
 
 const desc = `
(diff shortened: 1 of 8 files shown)


cmd/telemeter-server/main.go
```diff
@@ -95,10 +96,10 @@ func main() {
 	cmd.Flags().StringVar(&opt.TLSKeyPath, "tls-key", opt.TLSKeyPath, "Path to a private key to serve TLS for external traffic.")
 	cmd.Flags().StringVar(&opt.TLSCertificatePath, "tls-crt", opt.TLSCertificatePath, "Path to a certificate to serve TLS for external traffic.")
 
-	cmd.Flags().StringArrayVar(&opt.LabelFlag, "label", opt.LabelFlag, "Labels to add to each outgoing metric, in key=value form.")
+	cmd.Flags().StringSliceVar(&opt.LabelFlag, "label", opt.LabelFlag, "Labels to add to each outgoing metric, in key=value form.")
 	cmd.Flags().StringVar(&opt.PartitionKey, "partition-label", opt.PartitionKey, "The label to separate incoming data on. This label will be required for callers to include.")
 
-	cmd.Flags().StringArrayVar(&opt.Members, "join", opt.Members, "One or more host:ports to contact to find other peers.")
+	cmd.Flags().StringSliceVar(&opt.Members, "join", opt.Members, "One or more host:ports to contact to find other peers.")
```
squat: changed to `StringSliceVar` so a list of comma-separated value is parsed correctly.

pkg/authorize/jwt/handler.go
```diff
@@ -95,6 +95,9 @@ func (a *authorizeClusterHandler) ServeHTTP(w http.ResponseWriter, req *http.Req
 	labels := map[string]string{
 		a.partitionKey: cluster,
 	}
+	for k, v := range a.labels {
+		labels[k] = v
```
squat: This is the crux of the entire pull request

pkg/http/server/server.go
```diff
@@ -15,33 +15,33 @@ import (
 	"github.com/openshift/telemeter/pkg/store"
 	"github.com/openshift/telemeter/pkg/store/instrumented"
 	"github.com/openshift/telemeter/pkg/store/ratelimited"
+	"github.com/openshift/telemeter/pkg/validate"
 )
 
-type UploadValidator interface {
```
squat: This did not really belong here IMHO, but rather the validate package

pkg/metricfamily/multi_transformer.go
```diff
@@ -10,7 +10,9 @@ type MultiTransformer struct {
 }
 
 func (a *MultiTransformer) With(t Transformer) {
-	a.transformers = append(a.transformers, t)
+	if t != nil {
```
squat: a sanity check so we don't add nil-valued interfaces

cmd/telemeter-server/main.go
s-urbaniak: :+1:

PR #88, opened 2018-11-12
jsonnet: use internal endpoint for telemeter servicemonitors

s-urbaniak:
cc @squat


s-urbaniak:
@squat can you re-`/lgtm`?, prow didn't seem to pick it up.

PR #91, opened 2018-11-12
 jsonnet/telemeter: use TLS for telemeter server

squat:
This PR turns on TLS for the Telemeter server internal HTTP servers.
In order to enable this change, the Telemeter Prometheus instance must
be configured to load the serving service CA bundle. Furthermore, the
ServiceMonitors must specify this CA bundle in their TLSConfigs.

This PR also restructures the jsonnet/telemeter directory. The intention
here is to promote easier reuse of shared functions between the
different components. Chiefly, this allows the Telemeter server
libsonnet to reuse the `withNamespace` function. The reason for this
is that we need to substitute the namespace into the ServiceMonitor TLS
config `serverName` fields so that the field matches the SAN of the
injected service serving certificate and Prometheus is able to scrape
the target.

cc @s-urbaniak @brancz 
@riuvshin this PR renames some of the variables for the manifest templates to make them more unique and also introduces the `NAMESPACE` variable for the Telemeter server. Will we need to modify the staging CI configuration?


s-urbaniak:
no concerns from my side :+1: but let's wait for @riuvshin to sync up the CI configuration.

manifests/prometheus/list.yaml
```diff
@@ -246,19 +248,28 @@ objects:
...
-        serverName: prometheus-telemeter.telemeter.svc
+        caFile: /etc/prometheus/configmaps/prometheus-serving-certs-ca-bundle/service-ca.crt
+        serverName: prometheus-telemeter.${NAMESPACE}.svc
     selector:
       matchLabels:
         prometheus: telemeter
+- apiVersion: v1
+  data:
+    service-ca.crt: ""
+  kind: ConfigMap
```
riuvshin: ConfigMap are filtered by saasherder so we will need to create it manually.

manifests/prometheus/servingCertsCABundle.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+data:
+  service-ca.crt: ""
+kind: ConfigMap
```
riuvshin: is this exactly the same CM that is in the manifests/prometheus/list.yaml?

squat: Yes. All Prometheus manifests are written to `manifests/prometheus/` and one templateable list manifest is written to `manifests/prometheus/list.yaml`. This list includes all of the manifests. Same goes for the Telemeter server.

riuvshin: I see, thanks for explanation

jsonnet/telemeter/server/kubernetes/kubernetes.libsonnet
```diff
@@ -160,7 +161,11 @@ local clusterPort = 8082;
               bearerTokenFile: '/var/run/secrets/kubernetes.io/serviceaccount/token',
               interval: '30s',
               port: 'internal',
-              scheme: 'http',
+              scheme: 'https',
+              tlsConfig: {
+                caFile: '/etc/prometheus/configmaps/prometheus-serving-certs-ca-bundle/service-ca.crt',
```
s-urbaniak: the prod cluster is having a 3.11 openshift version. Shouldn't this be then the old location for service certs?

squat: hmm very true. We actually have a deeper issue here. Master is intended for 4.0, whereas anything destined for 3.11 should go in the release-3.11 branch. Production won't always be a 3.11 cluster. We really need to fix our deployment strategy and potentially carry patches to the 3.11 branch so we can correctly deploy the server on 3.11 AND 4.0 clusters.

squat: fixed for 3.11 now :)

PR #96, opened 2018-11-15
cmd/telemeter-server: improve rate limiting

squat:
Wrap the cluster store within a rate-limited store.
This guarantees an upper-bound on the total
inter-node requests that hit the target node of `l*n`, where l is the
rate limit and n is the cluster size. Without this, if a DOS attack
with IDs that hash to node A's bucket enter the cluster on different
node, node B, then node B will dutifully pass along the requests to
the node A and can DOS the target and congest the internal network.

cc @s-urbaniak @brancz


brancz:
Looks good, giving @s-urbaniak the last call.

PR #101, opened 2018-11-30
cmd,pkg,manifests: add whitelist option to server

squat:
This commit enabled whitelisting of metrics on the Telemeter server.
The whitelist can be configured by passing the `--whitelist` and
`--whitelist-file` flags. Every whitelist item (and every row of the
whitelist file) must be a valid PromQL match rule, e.g.
`{__name__="up",foo="bar"}`.

cc @brancz @s-urbaniak


pkg/http/server/server.go
```diff
@@ -159,18 +159,9 @@ func (s *Server) decodeAndStoreMetrics(ctx context.Context, partitionKey string,
 		return err
 	}
 
-	// filter the list
-	for i, family := range families {
```
squat: this was an exact re-write of the Filter transform, so let's substitute for conciseness.

pkg/metricfamily/required.go
```diff
@@ -15,8 +15,7 @@ func NewRequiredLabels(labels map[string]string) Transformer {
 }
 
 var (
-	ErrRequiredLabelValueMismatch = fmt.Errorf("a label value does not match the required value")
```
squat: This was unused

pkg/metricfamily/whitelist.go
```diff
@@ -0,0 +1,63 @@
...
+				continue Metric
+			}
+		}
+		family.Metric[i] = nil
+	}
+	return keep, nil
+}
+
+// match checks whether every Matcher matches a given metric.
+func match(name string, metric *clientmodel.Metric, matchers ...*labels.Matcher) bool {
```
squat: @brancz is there a more semantic way to implement this match check against a family using a Prometheus library? The only other _nice_ option I though of was loading the metric into a temporary tsdb storage struct so I could use the storage.Querier interface, however that seemed a little heavy duty for this simple check.

brancz: That is definitely too heavy, also it wouldn’t really satisfy the interface we’re looking for here. I would just suggest a function that takes a labelset and a slice of matchers, and each label of the labelset must pass every marcher. If we make sure the matchers (including Regex matchers) are compiles at start time of telemeter we should be good.

cmd/telemeter-server/main.go
```diff
@@ -382,9 +408,10 @@ func (o *Options) Run() error {
 		internalProtected.Handle("/debug/cluster", c)
 	}
 
-	var transforms metricfamily.Transformer
+	transforms := metricfamily.MultiTransformer{}
+	transforms.With(whitelister)
 	if len(o.Labels) > 0 {
-		transforms = metricfamily.NewLabel(o.Labels, nil)
+		transforms.With(metricfamily.NewLabel(o.Labels, nil))
```
s-urbaniak: Unrelated and not to ourselves: as far as I see, the Label transformer is modifying its state at every http request in an unsynchronized way:

https://github.com/openshift/telemeter/blob/6d3dc6367a168fce7e5a9ea9f1a43240c7f2dc96/pkg/metricfamily/label.go#L36

We should synchronize those modifications in the label transformer.

pkg/http/server/server.go
s-urbaniak: :+1:

pkg/metricfamily/whitelist.go
s-urbaniak: The underlying type of a `LabelSet` is a `map[LabelName]LabelValue`:

https://github.com/prometheus/common/blob/4724e9255275ce38f7179b2478abeae4e28c904f/model/labelset.go#L28

I do agree that the signature would look a little cleaner, but the current code is a little bit more efficient. The current code also already ensures compilation in the constructor.

PR #102, opened 2018-12-03
Fix ca bundle used to talk to telemeter client for metrics

brancz:
I was checking out latest cluster-monitoring-operator on OpenShift, and saw that telemeter was not successfully being scraped with a certificate error:

screenshot from 2018-12-03 17-33-08

I dug and found that kube-rbac-proxy was showing errors on requests (just an excerpt, but all log lines ever logged look like this):

```
2018/12/03 16:31:42 http: TLS handshake error from 10.129.0.12:48340: remote error: tls: bad certificate
2018/12/03 16:31:46 http: TLS handshake error from 10.128.0.39:43970: remote error: tls: bad certificate
2018/12/03 16:32:12 http: TLS handshake error from 10.129.0.12:48514: remote error: tls: bad certificate
```

So I looked at the config and found this subtle difference between the other `ServiceMonitor`s using the `service-ca.crt`.

@s-urbaniak @squat @metalmatze

### DIFF
--- a/jsonnet/telemeter/client/kubernetes.libsonnet
+++ b/jsonnet/telemeter/client/kubernetes.libsonnet
@@ -208,7 +208,7 @@ local securePort = 8443;
               port: 'https',
               scheme: 'https',
               tlsConfig: {
-                caFile: '/etc/prometheus/configmaps/telemeter-client-serving-certs-ca-bundle/%s' % servingCertsCABundleFileName,
+                caFile: '/etc/prometheus/configmaps/serving-certs-ca-bundle/%s' % servingCertsCABundleFileName,
                 serverName: $._config.telemeterClient.serverName,
               },
             },


jsonnet/telemeter/client/kubernetes.libsonnet
```diff
@@ -208,7 +208,7 @@ local securePort = 8443;
               port: 'https',
               scheme: 'https',
               tlsConfig: {
-                caFile: '/etc/prometheus/configmaps/telemeter-client-serving-certs-ca-bundle/%s' % servingCertsCABundleFileName,
```
squat: hmm but we have a configmap with that name: https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/telemeter-client/serving-certs-c-a-bundle.yaml#L8. Maybe the issue is that we simply need to tadd that configmap to the prometheus-k8s manifest so it is loaded in the statefusets

brancz: but it's the same ca for all serving certs, why multiple mounts then?

squat: yes this is true these bundle's will be the same. I think there is some justification to having each component be able to ship all of the dependencies needed to use/interact with it: "I am prometheus/I am telemeter and these are the CA's you need if you want to interact with me". Maybe the relationship with the CA bundle should be the inverse, anything that needs to interact with a service needs to provision its own bundle, in which case we should have a "cluster-monitoring-serving-certs-ca-bundle".

brancz: Yeah I can see both sides as well, my biggest argument (currently) for having a single cert (probably even more generic along `serving-certs-ca-bundle`, is what I'm thinking), is because the control of mounting additional CAs is on the Prometheus side. I'm happy to rename the mount, I would start in this PR and adapt everywhere in the PR that would integrate this into the cluster-monitoring-operator. Sounds good?

squat: sounds great. let's simplify

PR #103, opened 2018-12-04
pkg/metricfamily: lock label transformer

squat:
This commit introduces a mutex lock in the label transformer to ensure
that mutable state is not accessed concurrently.

cc @s-urbaniak addresses your comment https://github.com/openshift/telemeter/pull/101#discussion_r238182816

### DIFF
--- a/pkg/metricfamily/label.go
+++ b/pkg/metricfamily/label.go
@@ -1,6 +1,10 @@
 package metricfamily
 
-import clientmodel "github.com/prometheus/client_model/go"
+import (
+	"sync"
+
+	clientmodel "github.com/prometheus/client_model/go"
+)
 
 type LabelRetriever interface {
 	Labels() (map[string]string, error)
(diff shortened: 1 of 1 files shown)


pkg/metricfamily/label.go
```diff
@@ -9,6 +13,7 @@ type LabelRetriever interface {
 type label struct {
 	labels    map[string]*clientmodel.LabelPair
 	retriever LabelRetriever
+	mu        *sync.Mutex
```
s-urbaniak: can you leave a small comment why this must be a pointer?

squat: it doesn't _have_ to be

s-urbaniak: Ah, I was assuming it must be because of usages of the `label` struct as values rather than as references. In that case, actually a nit, I suggest to use a value instead of a pointer, it seems more idiomatic. But leaving the final call to you :-)

squat: Yeah, that was actually my first reaction as well, but then saw that the label struct is actually only referenced by reference and not value. I'll update it

PR #104, opened 2018-12-06
Add comments to each of the metrics in telemeter to explain there use

smarterclayton:
Ensure we have clear documentation for why these metrics should be captured
and what their goal is


squat:
Thanks Clayton. Can you fix the symlink at the root of the repo for metrics.json? (That file is there just to help surface the metrics and put them front and center)

jsonnet/telemeter/metrics.jsonnet
```diff
@@ -0,0 +1,48 @@
+[
+    // up contains information relevant to the health of the registered
+    // cluster monitoring sources on a cluster. This metric allows telemetry
+    // identify when an update causes a service to begin to crashloop or flake.
+    "{__name__=\"up\"}",
```
squat: Now that this is jsonnet and not strict json,I can make a follow up to clean up the quotes eg `'{__name__="up"}'`

jsonnet/telemeter/metrics.jsonnet
```diff
@@ -0,0 +1,48 @@
+[
+    // up contains information relevant to the health of the registered
+    // cluster monitoring sources on a cluster. This metric allows telemetry
+    // identify when an update causes a service to begin to crashloop or flake.
```
squat: Nit: missing “to” between “telemetry” and “identify”

smarterclayton:
Updated

PR #107, opened 2018-12-11
 jsonnet: fix prometheus SAR

squat:
This commit fixes the definition of the SAR for the proxy sitting in
front of the Telemeter Prometheus. The SAR was specifying an undefined
field `resourceName`, which was turning the SAR into `get namespace *`
rather than `get namespace $(NAMESPACE)`, which would fail.

This commit also removes the `namespace` field from the SAR, since
namespaces themselves are not namespaced objects.

This is currently breaking programmatic access via ServiceAccount tokens
to the Telemeter Prometheus instance in production.

cc @s-urbaniak @pbergene

### DIFF
--- a/jsonnet/jsonnetfile.lock.json
+++ b/jsonnet/jsonnetfile.lock.json
@@ -8,7 +8,7 @@
                     "subdir": "jsonnet/telemeter"
                 }
             },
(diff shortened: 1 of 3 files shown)


s-urbaniak:
/lgtm

@squat duh, very good catch :+1:

PR #108, opened 2018-12-12
docs: document the collected metrics.

squat:
cc @s-urbaniak @kbsingh


squat:
I think it’s fine if it’s best effort for now as these are just examples at the end of th day. The only truth is the metrics matching rules. Creating the e2e tests and other tasks are still higher priority

PR #113, opened 2019-01-18
[WIP] Add metrics for _cloud.openshift.com_

jhernand:
This patch adds metrics that will be used to report use of CPU, memory
and storage in the _cloud.openshift.com_ portal.


jhernand:
@yaacov, @kbsingh, oourfali please review.

Note that the metric that we are requesting for disk usage is actually the percentage of disk bandwidth used, so we will probably need something different.

jhernand:
@mxinden I am not 100% sure if this is the only change that it needs to be done, please let us know if something else is needed.

jsonnet/telemeter/metrics.jsonnet
```diff
@@ -46,4 +46,14 @@
...
+  // node_cpu_utilisation is used to report the current use of CPU
+  // in the cloud.openshift.com portal.
+  '{__name__="node:node_cpu_utilisation:avg1m"}',
+  // node_memory_utilisation is used to report the current use of
+  // memory in the cloud.openshift.com portal.
+  '{__name__="node:node_memory_utilisation:ratio"}',
+  // node_filesystem_size_bytes and node_disk_utilisation are used to
+  // report the current of of storage in the cloud.openshift.com portal.
+  '{__name__="node_filesystem_size_bytes"}',
+  '{__name__="node:node_disk_utilisation:avg_irate"}',
```
jhernand: @yaacov I think that `node_disk_utilisation` is the utilisation of disk bandwidth, not the utilisation of disk space, so we will probably need a different metric.

mxinden:
> @mxinden I am not 100% sure if this is the only change that it needs to be done, please let us know if something else is needed.

@squat @s-urbaniak would one of you mind taking a look? I have not been involved with Telemeter so far.

jhernand:
Thanks @squat. There is no _generate_ target in the Makefile, but I did run just `make` and added a new commit with the changes it produced. Is that correct?

jsonnet/telemeter/metrics.jsonnet
yaacov: @jhernand , for free fs space we may use:
 '{__name__="node_filesystem_free_bytes"}'

My reason for choosing  bandwidth over space utilization is copy/paste from the grafana dashboard, I assume we will be compared to the grafana dashboard, and get BZ's on metrics that are not the same :-(

@mrsiano hi, can you help here ?

I want to keep the metrics as close as possible to the metrics shown on openshifts grafana dashboard, they use `node_disk_utilisation` , but as @jhernand mentioned above, this is the metrics for fs bandwidth and not for fs space.

Since we can not show all the metrics possible, and need to choose a very small subset of metrics to show, maybe it's better to show the fs bandwidth like the grafana dashboard ?

On the other hand, when using grafana dashboard user can see a lot more metrics, so maybe we need to choose some other metrics ?

p.s
I did not find the repository holding the grafana dashboard used by the origin console, anyone know where it is ?
It make sense to me, to keep an eye on what they are doing, so we can keep using the same metrics on our end it they update their dashboards ...

yaacov:
@jhernand Thanks for making the PR.

Since telemetry is our only source of data for (none hive) clusters, I was going to go over again, 
to look for metrics that can give us:
a - the number of workers and masters.
b - a url for the orogin console.
I looked several times, and did not find metrics for that missing info :-(

From my side, all the metrics we expose in the API are present:
cpu tatal/usage, memory total/usage, file system total/usage are present.

We are still missing: number of worker/master nodes and url of cluster, but I do not think we can get it from the clusters prometheus data.

LGTM except for the fs usage issue.

mrsiano:
@yaacov @jhernand 
if you completely want to sync the metrics collections.
you should follow the mixin-metrics this is how the operator generate the metrics from.

https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/dashboards

yaacov:
@mrsiano Thanks :+1:  I will probably have more questions offline :-)

yaacov:
p.s.
Now that @mrsiano :100: pointed us to the source, we can do some better decision making about the metrics we want.

p.p.s.
@jhernand , the dashboard I used as reference is the `use` one, (I looked at the grafana UI, looking at the source is more informative):
https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/dashboards/use.libsonnet

yaacov:
As @jhernand requested, continuing at a new PR, #114.

squat:
closing this in favor of #114

PR #114, opened 2019-01-21
Add metrics for _cloud.openshift.com_

yaacov:
[ Taking over PR #113 from @jhernand ]

This patch adds metrics that will be used for reports in the cloud.openshift.com portal.

Ref:
https://github.com/openshift/telemeter/pull/113


brancz:
Each of these have a cardinality of # of nodes and they're 5 of them (we currently have 2 of that type). We're going to more than double the traffic produced against telemeter by this. Are we sure we're prepared for this? cc @squat

yaacov:
@jhernand, @kbsingh, @oourfali please review.

This is a copy of #113 with `node_filesystem_free_bytes` instead of `node:node_disk_utilisation:avg_irate`

We are using metrics fields similar to the fields used by the Grafana dashboards.
Rational for using Grtafana as reference, is that we want Openshift users to see similar metrics in their Openshift console:

Ref:
https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/dashboards
https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/dashboards/use.libsonnet

squat:
I don't think there should be any issue ingesting this, however, we will certainly need to adjust our calculations for PVC size based on these metrics, as the number of time series will grow greatly due to this addition.

yaacov:
We only use the sum of this, e.g. `sum({__name__="node_filesystem_size_bytes"})`
can this help ?

yaacov:
@squat I do not know how to fix the failing test, can you help with the failing test ?

kbsingh:
ping @sichvoge

squat:
@yaacov it looks like a flake; the cluster failed to bootstrap and the tests timed-out waiting. I'll restart it.

sichvoge:
From a product / use case perspective this lgtm. As a summary, these metrics are a critical part on providing health information to our customers on the _cloud.openshift.com_ portal.

@yaacov please work with @squat on the concerns raised by him and @brancz.

yaacov:
> @yaacov please work with @squat on the concerns raised by him and @brancz.

@sichvoge sure :+1: , the best thing I can think of is using `sum`, so we get one metrics per cluster, instead of metric per machine, a - is it possible, will is lose needed info (e.g. labels) ? b - does that makes sense from your side ( e.g. will it help with storage size ) ?

yaacov:
@brancz @squat we  [ currently :-) ]  do not show drill down views, only cluster metrics, so summing up is fine from my end.

please re-review.

yaacov:
@jhernand is it ok from our side to have only sums (e.g. cluster metrics, and loose the per node data and labels) ?

jsonnet/telemeter/metrics.jsonnet
```diff
@@ -46,4 +46,16 @@
   // cluster alerting for the purposes of preventing upgrades from
   // failing for end users.
   '{__name__="ALERTS",alertstate="firing"}',
+  // node_cpu_utilisation is used to report the current avg use of CPU
+  // in the cloud.openshift.com portal.
+  'sum({__name__="node:node_cpu_utilisation:avg1m"})',
+  // node_memory_utilisation is used to report the current use of
+  // memory in the cloud.openshift.com portal.
+  'sum({__name__="node:node_memory_utilisation:ratio"})',
+  // node_filesystem_free_bytes is used to
```
squat: This comment is referencing the wrong metric. It should say `node_filesystem_size_bytes`

sichvoge:
We are currently investigating options to be able to replicate more effectively, but we need to live with some restrictions right now.

jhernand:
@yaacov It is OK to have aggregated cluster metrics, yes.

jsonnet/telemeter/metrics.jsonnet
yaacov: Sorry, fixing.

yaacov:
@squat thanks for the check.

p.s. squashing.

yaacov:
@squat hi, I fixed the wrong docstring.

I need help with the `sum` missing labels, will this work in "real life" will the metrics federated have a label to know what they are ?

If the labels will be missing in federation, we can not do the `sum` aggregation :-(

yaacov:
@squat hi,

Tried to check the metrics generated when having an aggravation function ( e.g. sum ) by:
```
make
./test/integration.sh http://localhost:9005
```
and then going to http://localhost:9005/graph

I didn't see any relevant metrics, what am I missing ?

yaacov:
@squat sorry I didn't see the discussion about this :sob: , opend a PR in  cluster-monitoring-operator

https://github.com/openshift/cluster-monitoring-operator/pull/220

yaacov:
We do not yet know the name of the aggregated fields, waiting for https://github.com/openshift/cluster-monitoring-operator/pull/220.

kbsingh:
In the interest of making progress, can we close this issue collecting these metrics per node as was the original ask - and then raise a seperate thread around options to create new items in cluser / in transit and in central telemeter ?

yaacov:
>  can we close this issue collecting these metrics per node as was the original ask ?

Reverted to collecting per node. @squat PTAL

yaacov:
@squat heads up, filesystem metrics adds a lot of metrics, do you have a better metrics ?

brancz:
Both of the metrics added here have the cardinality of the number of nodes in a cluster, that's not acceptable as we can't plan well for a capacity like that. For our purpose, total CPU seconds and total memory used are sufficient aggregated across the cluster.

jhernand:
@brancz Are there already metrics for total CPU seconds and total memory? If so, what are the names?

yaacov:
closing in favour of #130

PR #115, opened 2019-01-21
Required metrics for  predicting service upgrade failures

SriKrishnaPaparaju:
These are the initial list of Kube resource specific metrics required for capturing the patterns of containers status (running, terminated, reason) to replay back at Canary OCP clusters before rolling out an upgrade.


kbsingh:
ping @sichvoge

kbsingh:
are these per container or a summary of status ?

squat:
@SriKrishnaPaparaju the generated files seem to be incorrect. can you please run `make` again and commit the result?

docs/data-collection.md
```diff
@@ -54,6 +54,14 @@ For the OpenShift 4 Developer Preview we will be sending back these exact attrib
   // cluster alerting for the purposes of preventing upgrades from
   // failing for end users.
   '{__name__="ALERTS",alertstate="firing"}',
+  // Kube state metrics for container status
+   '{__name__="kube_pod_container_status_running"}',
```
squat: These docs are still incorrect (you can see the indentation is off); can you please re-generate this file and commit it?

sichvoge:
@SriKrishnaPaparaju Can you remind me again what the flow for the canary cluster would be and how the different metrics will support a successful upgrade?

It would also be good to understand better, inline as a comment on the jsonnet file, how each metric will be used. So far, you only provide info about what the metric is about. We should always keep in mind that we point everyone to this list when they want to understand for what purpose we use certain metrics.

SriKrishnaPaparaju:
@sichvoge Please check this to understand the background for the metrics ask and overall plan for the roll out automation: https://docs.google.com/document/d/1SaBdSuPF1clmgqf3sK8jdA6yFOJoIM1YIEZzJ6hnjlM/edit#heading=h.pg8cntwd71qa

SriKrishnaPaparaju:
@kbsingh for the MVP, would it be possible to narrow down the scope the Kube-state metrics to two OpenShift namespaces only : 'openshift-console' and  'openshift-ingress' ?

squat:
@SriKrishnaPaparaju this will be a huge amount of time-series for every cluster. Can the same result be achieved using aggregated metrics instead? For example, a ratio of containers in different states plus the state of specific high-value containers like control plane components?

Second concern, does the team have an exception for this change given that code-freeze has passed?

brancz:
This needs a rebase. Could you also open a PR against the cluster-monitoring-operator to add these recording rules? They would go here: https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/rules.jsonnet (you can just follow the scheme of the kubernetes recording rules we already have there, and make sure to have [jsonnet installed](https://github.com/google/go-jsonnet#build-instructions) before running `make pkg/manifests/bindata.go`)

brancz:
Is this still relevant?

squat:
closing as obsolete

PR #116, opened 2019-01-22
 jsonnet/telemeter: bump rate-limit and replicas

squat:
Telemeter Prometheus is showing holes in the graphs due to timeouts
while Prometheus is scraping Telemeter server, where it is taking
\>10s on occasion. This commit scales the replicas so we can distribute
the metrics across the hash ring and rendering the metrics takes less
time. It also bumps the rate-limit so platforms that report metrics
on behalf of clusters, e.g. CI won't get rate-limited.

@jfchevrette @smarterclayton @brancz


squat:
https://github.com/openshift/telemeter/blob/master/jsonnet/telemeter/metrics.jsonnet#L48

PR #124, opened 2019-02-07
Add storage to prometheus

jfchevrette:
This adds 500Gi PVCs to prometheus replicas.

Commit contains both the jsonnet changes and the resulting manifest changes


jfchevrette:
Reopened at #126 to hopefully work around the prow/e2e test issues

PR #129, opened 2019-02-18
metrics: Gather the `cluster_version_available_updates` metric

smarterclayton:
This reports the upstream, the channel, and how many updates are
visible, which is critical to knowing whether clusters are seeing
the updates that they are entitled to have.


smarterclayton:
Hrm, has this not been rolled out to prod yet?

squat:
No, not yet. We simply need to bump the saas-herder repo. I’ll take care of it.

PR #130, opened 2019-02-27
This patch adds metrics that will be used to report use of CPU, memory

nimrodshn:
A rebase of #114 onto master.

cc: @jhernand @yaacov @kbsingh @oourfali @brancz @squat


nimrodshn:
@brancz @mxinden I am not quit sure why `ci/prow/generate` is failing can you help me?

squat:
@nimrodshn the generate test is failing because the generated secret values do not include `{__name__="cluster_version_available_updates"}`. It seems that you didn't regenerate the manifests after rebasing on https://github.com/openshift/telemeter/pull/129. Can you please rebase and re-generate.

Separately, as mentioned in https://github.com/openshift/telemeter/pull/114#issuecomment-467521346, we do not want to add these metrics like this, but as aggregated metrics. Are these the aggregated recording rule names?

brancz:
I agreed with kB that we’re taking this over.

nimrodshn:
Closing as there is no intention to add these metrics.

Issue #121, opened 2019-01-29
Move telemeter server metrics whitelist to a ConfigMap


jfchevrette:
https://github.com/openshift/telemeter/blob/ee4a19a2ba06b0c3d6f3bd471d90b7c6318ebfcf/manifests/server/list.yaml#L12

Secrets are ignored in upstream manifests and are instead fetched from a secrets store and applied during CI/CD. Because of this, the whitelist key in the above secret will never be applied to the running telemeter instances when changes are merged to this repo.

The established pattern is that non-secret / volatile data should reside in a ConfigMap which is not ignored by our CI which would allow changes to be merged without the intervention of the SRE team to update/apply a secret.

Would it be posible to move the whitelist to a ConfigMap object?

squat:
closed by #131

PR #131, opened 2019-02-28
 jsonnet/telemeter/server: move whitelist out of secret

squat:
This commit removes the whitelist from the Telemeter server secret and
places it instead in a series of command line `--whitelist` flags. This
is helpful for several reasons:
1. as the whitelist is no longer in a secret it no longer has to be
handled manually, preventing potential human error when the whitelist
changes; and
2. when the whitelist changes, the DaemonSet spec will also change so
the new configuration will automatically be rolled out without needing
to manually kick pods.

cc @s-urbaniak @jfchevrette


jsonnet/telemeter/server/kubernetes.libsonnet
```diff
@@ -50,9 +48,13 @@ local whitelistFileName = 'whitelist';
       local rhdUsername = containerEnv.fromSecretRef('RHD_USERNAME', secretName, 'rhd.username');
       local rhdPassword = containerEnv.fromSecretRef('RHD_PASSWORD', secretName, 'rhd.password');
       local rhdClientID = containerEnv.fromSecretRef('RHD_CLIENT_ID', secretName, 'rhd.client_id');
-      local secretMount = containerVolumeMount.new(secretVolumeName, secretMountPath);
       local secretVolume = volume.fromSecret(secretVolumeName, secretName);
 
+      local whitelist = std.map(
+        function(rule) "--whitelist='%s'" % std.strReplace(rule, 'ALERTS', 'alerts'),
```
s-urbaniak: does it make sense to do a std.asciiLower(str) here instead?

squat: no it does not make sense here; there may very well be metric names or labels in the whitelist that we do _not_ want completely lowercased. The only one we care about today is the `ALERTS` metric name.

s-urbaniak: agreed

squat:
fixes #121

PR #135, opened 2019-03-06
cmd/telemeter-server: instrument http clients

s-urbaniak:
This adds instrumentation of outgoing http requests towards both
tollbooth (authorize) endpoint and the configured oauth endpoint.

cc @brancz @squat


brancz:
top notch :+1: nice work

PR #136, opened 2019-03-06
added SLI metrics for openshift-sre

rafael-azevedo:
This change will add metrics that will be used in reports for SLI/SLOs for openshift-sre


jsonnet/telemeter/metrics.jsonnet
```diff
@@ -51,4 +51,8 @@
   // cluster alerting for the purposes of preventing upgrades from
   // failing for end users.
   '{__name__="ALERTS",alertstate="firing"}',
+  // the following three metrics will be used for SLA analysis reports.
```
brancz: please document the labels this metric can have, as it's not entirely obvious

jsonnet/telemeter/metrics.jsonnet
```diff
@@ -51,4 +51,8 @@
...
+  '{__name__="code:apiserver_request_count:rate:sum"}',
+  '{__name__="kube_pod_status_ready",condition="true",namespace="kube-system",pod=~"etcd.*"}',
```
brancz: if this is something we want to report, then it should be pre-aggregated in a recording rule, so we only ship the sum of pods in a state instead of a per pod cardinality

rafael-azevedo: I will make this change and put a PR into the monitoring operator

brancz:
/lgtm

even if the metrics don't exist in cluster-monitoring (yet), merging here first so we don't need to double update cluster-monitoring-operator

smarterclayton:
I am reverting this in #197  - these should be covered either by alerts or by the `up` metric, and if they aren't those are the appropriate places to fix those.

PR #137, opened 2019-03-07
add cluster instance type rules

tschuy:
Adds metrics from https://github.com/openshift/cluster-monitoring-operator/pull/281


squat:
closed by #143

PR #139, opened 2019-03-15
cmd/telemeter-server: start memory store cleaner

s-urbaniak:
Currently the cleaner is not started, causing possible memory leaks.

This fixes it.

cc @squat


brancz:
oops

/lgtm

PR #146, opened 2019-03-25
pkg/metricfamily: add elide transformer

s-urbaniak:
This adds the label eliding feature.
A separate PR follows which enables this feature in the deployment.

Fixes partially MON-516


s-urbaniak:
cc @squat

squat:
This PR looks good. But why implement it in Telemeter server and not just simply as a relabel config for Prometheus as discussed before?

s-urbaniak:
Honestly, no strong reasons, having all the manipulative operations in the deployment cli arguments was simply a low hanging fruit for me.

I can as well do the relabel config, wdyt?

s-urbaniak:
Some points also summarizing offline discussion with @squat :
- Consolidated mutation (transformers) package alongside with all the other manipulative metrics functions
- single-source-of-truth is still telemeter-server for collected metrics, if we move eliding to the prometheus scrape config, we have split mutation responsibilities for those metrics

I'll leave the final word to you @squat but given the low implementation complexity, I'd still leave this implementation in place.

(sorry for double posting, wrong tab)

PR #147, opened 2019-03-25
jsonnet/telemeter: elide prometheus_replica label

s-urbaniak:
Depends on #146 to be merged first

cc @squat

### DIFF
--- a/jsonnet/telemeter/server.libsonnet
+++ b/jsonnet/telemeter/server.libsonnet
@@ -15,6 +15,9 @@ local list = import 'lib/list.libsonnet';
     },
     telemeterServer+: {
       whitelist+: (import 'metrics.jsonnet'),
+      elideLabels+: [
+        'prometheus_replica',
+      ],
     },
   },
 }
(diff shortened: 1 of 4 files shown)


s-urbaniak:
Some points also summarizing offline discussion with @squat :
- Consolidated mutation (transformers) package alongside with all the other manipulative metrics functions
- single-source-of-truth is still telemeter-server for collected metrics, if we move eliding to the prometheus scrape config, we have split mutation responsibilities for those metrics

I'll leave the final word to you @squat but given the low implementation complexity, I'd still leave this implementation in place.

s-urbaniak:
bump @squat as #146 is merged

cben:
https://github.com/openshift/telemeter/blob/master/docs/sample-metrics.md and https://github.com/openshift/telemeter/blob/master/test/timeseries.txt still contain this label (valued `prometheus_replica="prometheus-k8s-0"` everywhere).
Do they represent data before eliding, or after and should be updated?

squat:
These are out of date and need to be updated :) thanks for pointing it out

PR #149, opened 2019-03-26
Add resource limits to telemeter server and prometheus

jfchevrette:
- Add resource requests and limits variables to telemeter server
- Add resource requests and limits variables to prometheus
- Allow setting the resource requests and limits via the templates

### DIFF
--- a/jsonnet/telemeter/server.libsonnet
+++ b/jsonnet/telemeter/server.libsonnet
@@ -19,14 +19,6 @@ local list = import 'lib/list.libsonnet';
       elideLabels+: [
         'prometheus_replica',
       ],
-      // resourceLimits: {
-      //   cpu: '1',
-      //   memory: '1Gi',
-      // },
-      // resourceRequests: {
-      //   cpu: '0.2',
-      //   memory: '100Mi',
-      // },
     },
   },
 }
--- a/jsonnet/telemeter/server.libsonnet
+++ b/jsonnet/telemeter/server.libsonnet
@@ -7,7 +7,7 @@ local list = import 'lib/list.libsonnet';
           + list.withAuthorizeURL($._config)
           + list.withNamespace($._config)
           + list.withServerImage($._config)
-          + list.withResourceRequestsAndLimits($._config)
+          + list.withResourceRequestsAndLimits('telemeter-server', $._config.telemeterServer.resourceRequests, $._config.telemeterServer.resourceLimits)
   },
 } + {
   _config+:: {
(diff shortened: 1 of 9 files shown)


derekwaynecarr:
It’s unclear if this is happening from my review, but please do not set a cpu limit.  Cpu requests should be sufficient.

s-urbaniak:
@derekwaynecarr the limits configured here do not refer to components running inside cluster installations. This configuration rather refers to components running on our premises (telemeter backend), supervisioned by service delivery. According to @jfchevrette limits are needed here for capacity management.

jfchevrette:
@derekwaynecarr can you elaborate on why you recommend not setting a cpu limit? 

From an SRE point of view, we find limits valuable as this allows us to control how much services can consume and prevent cluster saturation and alleviate the 'bad neighbor' effect. There is much less guesswork involved in capacity planning when we have cpu and memory limits on all services we run.

jsonnet/telemeter/server.libsonnet
```diff
@@ -18,6 +19,14 @@ local list = import 'lib/list.libsonnet';
       elideLabels+: [
         'prometheus_replica',
       ],
+      // resourceLimits: {
```
s-urbaniak: do you mind to remove those comments?

jfchevrette: removed in 2032d64 :+1:

manifests/prometheus/prometheus.yaml
```diff
@@ -43,6 +43,7 @@ spec:
     beta.kubernetes.io/os: linux
   replicas: 2
   resources:
+    limits: {}
```
s-urbaniak: @jfchevrette i didn't verify, but is this fine?

manifests/server/statefulSet.yaml
```diff
@@ -99,6 +99,9 @@ spec:
             path: /healthz/ready
             port: 8443
             scheme: HTTPS
+        resources:
+          limits: {}
+          requests: {}
```
s-urbaniak: see above

manifests/prometheus/prometheus.yaml
jfchevrette: I've tested that empty map is valid and the end result is that no limits are set. This is the same as before.

manifests/server/statefulSet.yaml
jfchevrette: I've tested that empty map is valid and the end result is that no limits are set. This is the same as before.

PR #152, opened 2019-04-02
authorize/jwt/handler: log authorize failures for >= 500

s-urbaniak:
Currently we still log all authorize failures, spamming the telemeter server logs.

This fixes it by reducing the logging to actionable errors, namely with http status code >= 500.

cc @squat @aditya-konarde: I recognized in production, that we indeed still log authorize failure, even if clusters are unknown. This should lower the noise on telemeter server logs in production.

### DIFF
--- a/pkg/authorize/jwt/handler.go
+++ b/pkg/authorize/jwt/handler.go
@@ -77,7 +77,9 @@ func (a *authorizeClusterHandler) ServeHTTP(w http.ResponseWriter, req *http.Req
 		}
 
 		if scerr, ok := err.(statusCodeErr); ok {
-			log.Printf("error: unable to authorize request: %v", scerr)
+			if scerr.HTTPStatusCode() >= http.StatusInternalServerError {
+				log.Printf("error: unable to authorize request: %v", scerr)
+			}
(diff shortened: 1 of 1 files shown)


pkg/authorize/jwt/handler.go
```diff
@@ -77,7 +77,9 @@ func (a *authorizeClusterHandler) ServeHTTP(w http.ResponseWriter, req *http.Req
 		}
 
 		if scerr, ok := err.(statusCodeErr); ok {
-			log.Printf("error: unable to authorize request: %v", scerr)
+			if scerr.HTTPStatusCode() >= http.StatusInternalServerError {
+				log.Printf("error: unable to authorize request: %v", scerr)
```
squat: Hmmm we should *really* be using leveled logging imo and log all failures at least as infos and all 5xxs as errors. For now this is ok

s-urbaniak: There are different schools out there, the Go ecosystem mostly circles around Peter's principles: https://peter.bourgon.org/blog/2016/02/07/logging-v-instrumentation.html

But k8s goes a different route obviously.

PR #153, opened 2019-04-04
 jsonnet: prometheus route balance source IP

squat:
This commit changes the Telemeter Prometheus Route object to use the
client's source IP as the load balancing criteria. This will ensure that
all requests coming from the same client end up at the same Pod. #145
was intended to implement this feature but came up short because
OpenShift Routes do not connect directly to Services but rather to their
underlying Endpoints.

cc @jfchevrette @s-urbaniak


squat:
rebased now that #154 is in

PR #155, opened 2019-04-05
DO NOT MERGE: Demonstrate removal of pointers from prometheus client serialization and memory impact

smarterclayton:
The primary scale blocker in telemeter-server is inefficient memory allocation caused by the extensive
use of pointers in the Go protobuf model objects created from metrics.proto. An older issue described
this in the upstream client library, telemeter demonstrates why high scale services should not use the
naive generated objects from google/protobuf (which prefers consistency rather than performance):
https://github.com/prometheus/client_golang/issues/303

gogo/protobuf which is in wide use (etcd, kubernetes) creates more normal objects when slavish adherence
to the google/protobuf models are not appropriate.

This demonstrates the changes that would be necessary to fix the prometheus client (which isn't even
truly dependent on the nullability of fields, since empty label pairs and names are equally invalid).

In testing locally with a fast clients (benchmark --workers 50 --interval 500ms) this reduced allocations
significantly (1-2 orders of magnitude).  There's likely further wins past this point (snappy allocations
should be pooled) and probably also some mutex blocking, but this conservatively would increase
telemeter-server headroom by 2-5x.

There are a number of options to consider - a slimmed down prometheus client, an upstream contribution,
a temporary fork (the model is effectively static so we could live with this while upstream considers
a move). OpenCensus has not made much progress in the last year or so.


squat:
We can close this now. will no longer be an issue in v2 and it is no longer the limiting factor for v1.

PR #156, opened 2019-04-05
Remove two major sources of contention from telemeter

smarterclayton:
While running local workloads I tested for contention between locks.
There were two obvious contention points that could be removed easily
without risk:

   store: Don't hold the mutex while writing metrics

   We were holding the limiter lock while accessing the store, which is not
   necessary because the store is already safe and the limiter is already a
   synchronization point. Switch the lock to guard only the map.

   Removes a significant source of contention from the path.

and

   cluster: Avoid contention on write lock

   A significant source of contentions in the block profile was the 
   refreshRing call. Since reads dominate writes (100:1 or more) we should us
   a RWMutex. Also, we don't need to refresh inline, so switch to a background
   loop for refreshing.

   Removes a source of contention when handling incoming requests.

To test locally you need to set `runtime.SetBlockProfileRate(>0)` and then
use `go tool pprof -http : http://localhost:9004/debug/pprof/block` after
a few minutes of stress testing.

### DIFF
--- a/pkg/store/ratelimited/ratelimited.go
+++ b/pkg/store/ratelimited/ratelimited.go
@@ -43,19 +43,22 @@ func (s *lstore) writeMetrics(ctx context.Context, p *store.PartitionedMetrics, 
 		return nil
 	}
 
-	s.mu.Lock()
-	defer s.mu.Unlock()
-
-	limiter, ok := s.store[p.PartitionKey]
-
-	if !ok {
-		limiter = rate.NewLimiter(rate.Every(s.limit), 1)
-		s.store[p.PartitionKey] = limiter
-	}
-
-	if !limiter.AllowN(now, 1) {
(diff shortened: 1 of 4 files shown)


pkg/store/ratelimited/ratelimited.go
```diff
@@ -43,19 +43,22 @@ func (s *lstore) writeMetrics(ctx context.Context, p *store.PartitionedMetrics,
 		return nil
 	}
 
+	if limiter := s.limiter(p.PartitionKey); !limiter.AllowN(now, 1) {
```
s-urbaniak: just as a verification point: I rechecked if `AllowN` does any mutations internally, it does, but they are also guarded internally by a mutex: https://github.com/golang/time/blob/master/rate/rate.go#L288.

So this should be fine.

smarterclayton:
I’ll fix the test, it assumes inline refresh still

s-urbaniak:
/lgtm
thanks @smarterclayton! :100:

PR #162, opened 2019-04-15
metrics.json: reduce etcd object counts

squat:
This commit greatly reduces the cardinality of the etcd_object_counts
metric by switching it in favor the new instance:etcd_object_counts:sum
recording rule introduced in
openshift/cluster-monitoring-operator#323. This
metric has much lower cardinality while still giving us some useful
insight into the cluster.

Depends on: https://github.com/openshift/cluster-monitoring-operator/pull/323

cc @brancz @s-urbaniak @smarterclayton


smarterclayton:
Love it

PR #164, opened 2019-04-16
WIP pkg/benchmark: simulate faulty clusters

s-urbaniak:
/cc @squat

### DIFF
--- a/cmd/telemeter-benchmark/main.go
+++ b/cmd/telemeter-benchmark/main.go
@@ -30,9 +30,10 @@ type options struct {
 	ToToken     string
 	ToTokenFile string
 
-	Interval    time.Duration
-	MetricsFile string
-	Workers     int
+	Interval          time.Duration
+	MetricsFile       string
+	Workers           int
+	FaultyProbability float64
 }
 
 var opt options = options{
@@ -61,6 +62,7 @@ func main() {
 	cmd.Flags().DurationVar(&opt.Interval, "interval", opt.Interval, "The interval between scrapes. Prometheus returns the last 5 minutes of metrics when invoking the federation endpoint.")
 	cmd.Flags().StringVar(&opt.Listen, "listen", opt.Listen, "A host:port to listen on for health and metrics.")
 	cmd.Flags().IntVar(&opt.Workers, "workers", opt.Workers, "The number of workers to run in parallel.")
+	cmd.Flags().Float64Var(&opt.FaultyProbability, "faulty-probability", opt.FaultyProbability, "The probability a worker is a cluster, which sends faulty payloads.")
 
 	if err := cmd.Execute(); err != nil {
 		os.Exit(1)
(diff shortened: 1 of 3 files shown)


cmd/telemeter-benchmark/main.go
```diff
@@ -61,6 +62,7 @@ func main() {
 	cmd.Flags().DurationVar(&opt.Interval, "interval", opt.Interval, "The interval between scrapes. Prometheus returns the last 5 minutes of metrics when invoking the federation endpoint.")
 	cmd.Flags().StringVar(&opt.Listen, "listen", opt.Listen, "A host:port to listen on for health and metrics.")
 	cmd.Flags().IntVar(&opt.Workers, "workers", opt.Workers, "The number of workers to run in parallel.")
+	cmd.Flags().Float64Var(&opt.FaultyProbability, "faulty-probability", opt.FaultyProbability, "The probability a worker is a cluster, which sends faulty payloads.")
```
squat: This comment needs a tweak. Right now it says “clusters are things that send faulty payloads and this is the probability that a worker is a cluster”.

We want “The pobability that a worker is a cluster that sends faulty payloads.”

Or simpler yet “The probability that a worker sends faulty payloads.”

pkg/metricsclient/faulty_sender.go
```diff
@@ -0,0 +1,80 @@
...
+		defer func() {
+			if _, err := io.Copy(ioutil.Discard, resp.Body); err != nil {
+				log.Printf("error copying body: %v", err)
+			}
+			resp.Body.Close()
+		}()
+
+		switch resp.StatusCode {
+		case http.StatusOK:
+			gaugeRequestSend.WithLabelValues(c.metricsName, "200").Inc()
```
squat: For consistency we could return nil in this case and eliminate the return at the end of the anonymous func.

pkg/metricsclient/faulty_sender.go
```diff
@@ -0,0 +1,80 @@
...
+	return &FaultySender{
+		client:      client,
+		timeout:     timeout,
+		metricsName: metricsName,
+	}
+}
+
+func (c *FaultySender) Send(ctx context.Context, req *http.Request, families []*clientmodel.MetricFamily) error {
+	buf := &bytes.Buffer{}
+	if _, err := buf.Write([]byte{0xd, 0xe, 0xa, 0xd, 0xb, 0xe, 0xe, 0xf}); err != nil {
```
squat: Instead of making a buffer pointer and then writing the bytes to it, which can fail, why not create a new buffer directly from byte slice? https://golang.org/pkg/bytes/#NewBuffer

s-urbaniak:
@squat thanks for the review! i literally hacked this in 5 minutes :-D It served its purposes today but I'll address the comments as I believe it's a valuable tool for fire drills too.

squat:
closing stale PR

PR #170, opened 2019-05-15
jsonnet/telemeter/prometheus: remove route declaration

s-urbaniak:
Routes are maintained in a different repo:
https://gitlab.cee.redhat.com/service/app-interface/blob/master/resources/app-sre

@squat @aditya-konarde


squat:
/lgtm

It would be much nicer to keep all of this config in one place IMO

PR #171, opened 2019-05-15
prometheus-telemeter: Set sessionAffinity to ClientIP

aditya-konarde:
This is so that any client within the cluster talking to prometheus gets to exactly the same server as long as the server is up.

### DIFF
--- a/deploy/prometheus.yaml
+++ b/deploy/prometheus.yaml
@@ -16,6 +16,7 @@ items:
     - name: http
       port: 80
       targetPort: 9090
+    sessionAffinity: ClientIP
 - kind: StatefulSet
   apiVersion: apps/v1
   metadata:


aditya-konarde:
@s-urbaniak

aditya-konarde:
Looks like its already in: https://github.com/openshift/telemeter/blob/master/manifests/prometheus/service.yaml#L18

I was just looking at the incorrect directory.

squat:
@aditya-konarde note that session affinity on services is ignored by OpenShift Routes, so the line of code that you pointed to here https://github.com/openshift/telemeter/pull/171#issuecomment-492585347 doesn't help. We need to use an annotation on a Route object as shown here: https://github.com/openshift/telemeter/pull/153

aditya-konarde:
@squat thanks for the info, that makes it clearer why we need it on the Route. 

PS: I didn't know of this before.

PR #173, opened 2019-05-16
Bump Prometheus Operator to v0.30.0 in jsonnet

metalmatze:
We want to pin kube-prometheus in the cluster-monitoring-operator to use the stable `release-0.1` branch. That branch of kube-proemtheus however already uses v0.30.0 of the Prometheus Operator and therefore this needs to be bump in here, because jsonnet-bundler cannot handle (and shouldn't) multiple version in the same repo.

/cc @s-urbaniak @squat

### DIFF
--- a/jsonnet/telemeter/jsonnetfile.json
+++ b/jsonnet/telemeter/jsonnetfile.json
@@ -18,7 +18,7 @@
                     "subdir": "jsonnet/prometheus-operator"
                 }
             },
-            "version": "v0.29.0"
+            "version": "v0.30.0"
         }
     ]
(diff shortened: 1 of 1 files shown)


squat:
There were no code changes from bumping the PO version?

metalmatze:
> There were no code changes from bumping the PO version?

No this only changes the configuration, which is fine, and not the Prometheus Operator code itself in this repository.

metalmatze:
Sorry, by accident I opened this PR against `master` when it should be against `release-4.1`.

/hold cancel

PR #182, opened 2019-06-07
pkg/oauth2: consider session cancellation

s-urbaniak:
Currently, if telemeter assumes a refresh token is still valid, it will
treat errors from fetching an access token as fatal.

In reality, the authorization server might have reset our session and
invalidate the refresh token server-side despite still being valid.

In that case it returns a 400 status code.

This fixes it.

Fixes MON-690

/cc @brancz @squat @paulfantom

### DIFF
--- a/pkg/oauth2/token_source.go
+++ b/pkg/oauth2/token_source.go
@@ -3,6 +3,7 @@ package oauth2
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"sync"
 	"time"
 
@@ -42,10 +43,6 @@ func NewPasswordCredentialsTokenSource(ctx context.Context, cfg *oauth2.Config, 
 }
 
 func (c *passwordCredentialsTokenSource) Token() (*oauth2.Token, error) {
-	return c.token(time.Now)
-}
-
-func (c *passwordCredentialsTokenSource) token(now func() time.Time) (*oauth2.Token, error) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
@@ -56,6 +53,12 @@ func (c *passwordCredentialsTokenSource) token(now func() time.Time) (*oauth2.To
 
 	if c.refreshToken.Valid() {
 		tok, err = c.accessTokenSource.Token()
+
+		rerr, ok := err.(*oauth2.RetrieveError)
+		if ok && rerr.Response != nil && rerr.Response.StatusCode == http.StatusBadRequest {
+			return c.passwordCredentialsToken()
+		}
+
 		if err != nil {
 			return nil, fmt.Errorf("access token source failed: %v", err)
 		}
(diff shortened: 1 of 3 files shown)


pkg/oauth2/token_source.go
```diff
@@ -65,26 +68,46 @@ func (c *passwordCredentialsTokenSource) token(now func() time.Time) (*oauth2.To
...
 		}
 
-		c.accessTokenSource = c.cfg.TokenSource(c.ctx, tok)
+		return tok, nil
 	}
 
+	return c.passwordCredentialsToken()
+}
+
+func (c *passwordCredentialsTokenSource) passwordCredentialsToken() (*oauth2.Token, error) {
```
brancz: This shouldn’t happen too often, but it would be good to know when it does. Let’s add a counter metric to understand how often this happens.

s-urbaniak: agreed :+1: added a metric, PTAL

pkg/oauth2/token_source.go
```diff
@@ -7,9 +7,23 @@ import (
...
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/oauth2"
 )
 
+var (
+	grantsTotal = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "password_credentials_grants_total",
```
brancz: this should have a `telemeter_` prefix

pkg/oauth2/token_source.go
```diff
@@ -7,9 +7,23 @@ import (
...
+			Help: "Tracks the number of resource owner password credential grants.",
+		},
+	)
+)
+
+func init() {
+	prometheus.MustRegister(grantsTotal)
```
brancz: Could we not use the global registry? I'd like us to generally move away from this pattern, and we should be a role model for the rest of the teams.

s-urbaniak: good idea. we can do a head-start here :+1: I suggest to dependency-inject the registry here, then gradually factor all the other places, wdyt?

brancz: that works for me

pkg/oauth2/token_source.go
s-urbaniak: fixed, ptal

pkg/oauth2/token_source.go
s-urbaniak: addressed, ptal

pkg/oauth2/token_source.go
```diff
@@ -65,26 +70,48 @@ func (c *passwordCredentialsTokenSource) token(now func() time.Time) (*oauth2.To
...
 
-		c.accessTokenSource = c.cfg.TokenSource(c.ctx, tok)
+		return tok, nil
+	}
+
+	return c.passwordCredentialsToken()
+}
+
+func (c *passwordCredentialsTokenSource) passwordCredentialsToken() (*oauth2.Token, error) {
+	c.grantsCounter()
```
brancz: it seems to me we may want to differentiate in a successful and errored token retrieval no?

pkg/oauth2/token_source.go
```diff
@@ -32,20 +34,17 @@ type passwordCredentialsTokenSource struct {
...
 	return &passwordCredentialsTokenSource{
-		ctx:      ctx,
-		username: username,
-		password: password,
-		cfg:      cfg,
+		ctx:           ctx,
+		username:      username,
+		password:      password,
+		cfg:           cfg,
+		grantsCounter: grantsCounter,
```
brancz: typically the actual metric is passed in here, or even grouped in a metrics struct

s-urbaniak: That was just to ease testing, but I can pass a metrics struct here (or the counter interface) too.

pkg/oauth2/token_source.go
s-urbaniak: We have that indication (status code returned by tollbooth) already in the `oauth` client metric, i.e. `client_api_requests_total{client="oauth",code="..."}`. This metric is also used to set up alerting.

Would it be idiomatic to add a `err="false|true"` here?

pkg/oauth2/token_source.go
s-urbaniak: adressed

pkg/oauth2/token_source.go
brancz: but it's differentiated in terms of refresh token use and credentials flow? if so lgtm

s-urbaniak: As discussed oob:
1. The existing `client_api_requests_total` metric does not differentiate the cause of the refresh. I added the `cause` label for that reason.
2. We'll add a `status` label here to have more insight on the result of the password grant operation.

ptal

PR #183, opened 2019-06-13
jsonnet: Bump prometheus-operator

brancz:
This is required to resolve the version conflict of bumping kube-prometheus in the cluster-monitoring-operator.

@s-urbaniak @squat @paulfantom @metalmatze

### DIFF
--- a/jsonnet/telemeter/jsonnetfile.json
+++ b/jsonnet/telemeter/jsonnetfile.json
@@ -18,7 +18,7 @@
                     "subdir": "jsonnet/prometheus-operator"
                 }
             },
-            "version": "v0.30.0"
+            "version": "release-0.30"
         }
     ]
 }
--- a/jsonnet/jsonnetfile.lock.json
+++ b/jsonnet/jsonnetfile.lock.json
@@ -8,7 +8,7 @@
                     "subdir": "jsonnet/telemeter"
                 }
             },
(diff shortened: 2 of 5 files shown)


brancz:
I realized what was the problem, we actually already had the v0.30.0 version (and I created this PR off of a very old state of master), but kube-prometheus specifies the `release-0.30` release series instead. Updated this PR accordingly.

jsonnet/telemeter/jsonnetfile.json
```diff
@@ -18,7 +18,7 @@
                     "subdir": "jsonnet/prometheus-operator"
                 }
             },
-            "version": "v0.30.0"
+            "version": "release-0.30"
```
s-urbaniak: good catch, but am I assuming correctly that we have to fix this also in the `release-4.1` branch?
https://github.com/openshift/telemeter/blob/c51e426c4ab344364b83cbdc5a76b096c13030b2/jsonnet/telemeter/jsonnetfile.json#L21

brancz: Yes, we should cherry-pick.

squat: it's already there. 4.1 was the first branch to be patched: https://github.com/openshift/telemeter/pull/173

brancz: the point is that both of these should be pointed at the minor release series, instead of the exact release

PR #186, opened 2019-06-18
pkg/store: fix ratelimited err overflow

squat:
cc @s-urbaniak @metalmatze @aditya-konarde


metalmatze:
/lgtm

Such a great find! So subtle. Very nice @squat!

PR #188, opened 2019-06-26
Forward to Thanos receive components

metalmatze:
This PR adds a new store, which is activated with `--forward-url` set to some URL.
The new store will forward every write request to a remote Thanos receive component.
If there's an error during forwarding, this error won't fail the v1 request. Instead it will still return `nil` and increase the `telemeter_forward_request_errors_total` counter.

Currently the unit tests are looking good and the metrics I can query in the Thanos qurier also look promising (correct labels and the value is set).
However, we should definitely make sure to add some end to end tests, with known and correct data as input and output!

I've added a histogram for RED metrics for the http request duration when forwarding and a errors total that counts the amount of transient forwarding errors. 
```
telemeter_forward_request_duration_seconds_bucket{status_code="200",le="0.005"} 35
telemeter_forward_request_duration_seconds_bucket{status_code="200",le="0.01"} 37
telemeter_forward_request_duration_seconds_bucket{status_code="200",le="0.025"} 38
telemeter_forward_request_duration_seconds_bucket{status_code="200",le="0.05"} 38
telemeter_forward_request_duration_seconds_bucket{status_code="200",le="0.1"} 38
telemeter_forward_request_duration_seconds_bucket{status_code="200",le="0.25"} 38
telemeter_forward_request_duration_seconds_bucket{status_code="200",le="0.5"} 38
telemeter_forward_request_duration_seconds_bucket{status_code="200",le="1"} 38
telemeter_forward_request_duration_seconds_bucket{status_code="200",le="2.5"} 38
telemeter_forward_request_duration_seconds_bucket{status_code="200",le="5"} 38
telemeter_forward_request_duration_seconds_bucket{status_code="200",le="+Inf"} 38
telemeter_forward_request_duration_seconds_sum{status_code="200"} 0.10843776099999998
telemeter_forward_request_duration_seconds_count{status_code="200"} 38
telemeter_forward_request_errors_total 0
```

/cc @squat @brancz


pkg/store/forward/forward.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/gogo/protobuf/proto"
+	"github.com/golang/snappy"
+	"github.com/openshift/telemeter/pkg/store"
+	clientmodel "github.com/prometheus/client_model/go"
+	"github.com/prometheus/prometheus/prompb"
+)
+
+const metricName = "__name__"
+
+type Store struct {
```
squat: rather than expose this type, let's make this unexported and return a `store.Store` from the `New` func. This way we can keep this tied to the store interface

pkg/store/forward/forward.go
```diff
@@ -0,0 +1,130 @@
...
+			return fmt.Errorf("response was not 200 OK, but %s", resp.Status)
+		}
+
+		return nil
+	}()
+	if err != nil {
+		log.Printf("forwarding error: %v", err)
+	}
+
+	return nil
```
squat: If this func will _always_ return a nil error, why not make the entire forwarding func async, that way we can write to memory sooner?

pkg/store/forward/forward.go
```diff
@@ -0,0 +1,130 @@
...
+
+			switch *f.Type {
+			case clientmodel.MetricType_COUNTER:
+				s.Value = *m.Counter.Value
+			case clientmodel.MetricType_GAUGE:
+				s.Value = *m.Gauge.Value
+			case clientmodel.MetricType_UNTYPED:
+				s.Value = *m.Untyped.Value
+			default:
+				panic(fmt.Sprintf("metric type %s not supported", f.Type.String()))
```
squat: Let's make this an error, yeah? otherwise, if there was a summary or something a panic and fail the entire WriteMetrics func and the metrics would never get written to memory either

pkg/store/forward/forward.go
```diff
@@ -11,13 +11,32 @@ import (
...
 const metricName = "__name__"
 
+var (
+	forwardErrors = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "telemeter_forward_request_errors_total",
+		Help: "Total amount of errors encountered while forwarding",
+	})
+	forwardDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "telemeter_forward_request_duration_seconds",
+		Help:    "Tracks the current amount of families for a given partition.",
```
squat: This seems to be copy-pasted

metalmatze: Indeed. I'll update.

pkg/store/forward/forward.go
metalmatze: I tried to follow the practice of always returning am exported concrete type, even when composing multiple types using the same interface. But I can update this.

pkg/store/forward/forward.go
metalmatze: Thought about making it async. But not totally sure about the implications of that. What do you think about keeping it like that and then move it to async later, if it becomes a problem?! :thinking:

pkg/store/forward/forward.go
metalmatze: Yes, you're right that's probably better. I'll refactor to trickle up the error.

pkg/store/forward/forward.go
squat: right now the implication of making it synchronous is that writing to memory will be delayed until writing to thanos has succeeded

metalmatze:
The forwarding happens in a separate goroutine now and thus unblocks the initial handler from waiting for the Thanos receive request to succeed.

pkg/store/forward/forward.go
```diff
@@ -0,0 +1,174 @@
...
+
+			return nil
+		}()
+		if err != nil {
+			forwardErrors.Inc()
+			log.Printf("forwarding error: %v", err)
+		}
+	}()
+
+	return nil
```
brancz: shouldn't we be writing to `next` at this point?

metalmatze: You're _totally_ right. I never tested if the data gets still into Telemeter :scream: :sweat_smile: Fixing!

brancz: That should be our acceptance criteria on staging.

PR #189, opened 2019-06-26
metrics: gather cluster_installer series

crawford:


squat:
@crawford do you have a running cluster you tested this on? if so, could you also run:
```sh
make test/timeseries.txt
make generate
```
to get the new metric in the `test/timeseries.txt` file and also generate an updated `docs/sample-metrics.md`?

If you don't have one on hand, don't worry, we can do a follow up :)

crawford:
@squat my generated timeseries data is _significantly_ different than the last. Is that a concern?

squat:
@crawford not an issue at all. It’s expected that pretty much every line should be different as timestamps and values will all change.

crawford:
@squat updated. `make generate` didn't produce any new assets. I'm not sure how to get that doc updated.

squat:
Sorry to complicate things. The makefile has recently changed. Plz try `make generate-in-docker -B` to force and build in a container.

crawford:
@squat Okay, `make generate -B` did the trick. Is there an existing process for approving what gets whitelisted?

sichvoge:
There is an internal approval chain for that. Currently, the monitoring team working on Telemeter and I approve new metrics that people like to push back. The engineering team makes sure that the metric is ok to send (specifically taking cardinality into account) and I just want to understand what the use cases are and if that alines with Telemeters goal.

TD;LR: you need two LGTM: 1) telemeter engineer and 2) me :)

squat:
/hold

From telemeter engineering team, this looks good to me. The new time series do not present a burden and are well within our current capacity.

sichvoge:
I am actually not really sure what the following means: 

> cluster_installer reports what installed the cluster, along with its version number and invoker.

What information are we sending off-cluster? Do you have an example?

squat:
Ah, that should have been captured in the file I asked @crawford to generate, `test/time-series.txt`. But I see now that it's not there. @crawford the metric does not appear to be in Prometheus. Is the service exposing this metric not yet being scraped?

crawford:
Hmm, let me take a closer look. I'll update the commit message with an example as well.

crawford:
@squat Okay, this should be ready for review now.

smarterclayton:
/approve

This metric is approved at a product level (high value in both triage and identification of unique issues)

crawford:
Rebased onto master. @squat can you take _one_ more look?

squat:
ahh we had another conflict

squat:
/lgtm
cmooooooooon

smarterclayton:
Does this also need to be added to the whitelist?

crawford:
@smarterclayton yes it does. Where did the server manifests go?

squat:
The telemeter-server manifests are changing home to their new v2 location in github.com/observatorium/configuration. The important thing here in this PR is to update the metrics.json file. This is the source of truth and will get picked up in the new v2 configuration repo (including the telemeter-server whitelist) the next time we bump.

crawford:
@squat cool. Do you know what's up with those two failed tests? Did I do something wrong in the PR or should I just kick them again?

squat:
@crawford let me dig

PR #191, opened 2019-07-09
Replaced code:apiserver_request_count:rate:sum metric with apiserver_request_count

abyrne55:
In order to accurately track SLAs/SLOs that use request counts as an indicator (SLI), we need to count the EXACT number of requests arriving over a time period using PromQL's "increase()" function. This function only operates properly on raw counter metrics, so the existing sum()'d and rate()'d recording rule will not suffice.


abyrne55:
I was informed that the cardinality of this metric is much too high to function as a recording rule, so I'm closing this and will be submitting an alternative rule PR shortly

PR #192, opened 2019-07-11
Added metrics list for OCS

anmolsachan:
Signed-off-by: Anmol Sachan <anmol13694@gmail.com>


brancz:
Assuming each of these have a cardinality of exactly 1 per cluster, that's still 16 time-series.

1) Can you confirm that each of these has exactly a cardinality of 1.
2) We haven't rolled out telemeter v2 yet, so we still have a very strict time-series limit, which of these would you disable if you had to?

Also these are intended to only exist starting 4.2 if I understand correctly?

anmolsachan:
@pcuzner @shtripat Please review the metrics and the respective comments.

anmolsachan:
> Assuming each of these have a cardinality of exactly 1 per cluster, that's still 16 time-series.
> 
> 1. Can you confirm that each of these has exactly a cardinality of 1.
I will re-check and update here 
> 2. We haven't rolled out telemeter v2 yet, so we still have a very strict time-series limit, which of these would you disable if you had to?
> 
I will check and update the same post some discussion soon.
> Also these are intended to only exist starting 4.2 if I understand correctly?
Yes, 4.2 it is targeted for.
@brancz Thanks for the pointers.

jsonnet/telemeter/metrics.jsonnet
```diff
@@ -73,4 +73,38 @@
   // cluster:node_instance_type_count:sum is the number of nodes
   // of each instance type and role.
   '{__name__="cluster:node_instance_type_count:sum"}',
+  //
+  //OCS metrics to be collected:
+  // job:ceph_osd_in:count gives the count of ceph osd which are in 'in' state.
+  '{__name__="job:ceph_osd_in:count"}',
```
smarterclayton: Unfortunately these are all incredibly specific - we need to have higher level summaries here.  This is a substantial portion of our metrics budget, so we can't add these as is.  We probably need to discuss in detail what the limitations are and how you can fit within a budget.

smarterclayton: Also, I'd like to see what alerts you have configured for this already if you can link that.

https://github.com/openshift/telemeter/pull/194/files is great - it's a single metric that shows us data about usage.  I asked whether there are a few other ones that show failure rates or whether they have alerts.  I would recommend starting by having 3-10 series that have the highest value input.

pcuzner: @smarterclayton understanding the budget we have to play with would have been nice ;). All the metrics we've identified are driven from a support angle - if we can't make them fit within the payload I have some ideas of where compromises could be made. However, there are another 6 metrics coming from Noobaa. Will discuss further with @anmolsachan and @shtripat

pcuzner: @smarterclayton at this point we have a list of about 11-12 that cover Ceph and Noobaa.

shtripat:
@brancz @smarterclayton we re-looked at the list of metrics and now its reduced to 7.
Kindly have a look. This is something we came up after a discussion within the team and bare minimum we would need this much.

anmolsachan:
> @brancz @smarterclayton we re-looked at the list of metrics and now its reduced to 7.
> Kindly have a look. This is something we came up after a discussion within the team and bare minimum we would need this much.

It's 8 actually, one additional for ceph versions.

smarterclayton:
Do you have an example I can see of these metrics on a live cluster (cut and pasted from a live cluster)?

anmolsachan:
> Do you have an example I can see of these metrics on a live cluster (cut and pasted from a live cluster)?

@smarterclayton Due to some upstream issues on Rook, I am unable to create a setup which can have these metrics. I will provide that asap.
The rules mentioned in PR come from recording rules in ceph-mixins : https://github.com/ceph/ceph-mixins/blob/e3f1fbdab5a5bec987a8e92e898254b3ea75d5d3/extras/manifests/prometheus-ceph-rules.yaml#L19-L35

anmolsachan:
Screenshots for the mentioned metrics in the PR:

Screenshot from 2019-07-24 15-32-53
Screenshot from 2019-07-24 15-33-03

anmolsachan:
@smarterclayton @squat I have Updated the PR according to the recommendations. Please review.

Added the screenshots in https://github.com/openshift/telemeter/pull/192#issuecomment-514568851

s-urbaniak:
/assign squat

also let's wait for final approval of @smarterclayton

docs/data-collection.md
```diff
@@ -84,6 +84,24 @@ For the OpenShift 4 Developer Preview we will be sending back these exact attrib
...
+  //OCS metrics to be collected:
+  // ceph_cluster_total_bytes gives the size of ceph cluster in bytes.
+  '{__name__="ceph_cluster_total_bytes"}',
+  // ceph_cluster_total_used_raw_bytes is the amount of ceph cluster storage used in bytes.
+  '{__name__="ceph_cluster_total_used_raw_bytes"}',
+  // ceph_health_status gives the ceph cluster health status
+  '{__name__="ceph_health_status"}',
+  // job:ceph_osd_metadata:count is the total count of osds.
+  '{__name__="job:ceph_osd_metadata:count"}',
+  // job:kube_pv:count is the total count pv's PV's present in OCP cluster.
```
squat: Please rephrase this description, I can’t quite understand what this metric accomplishes.

docs/data-collection.md
```diff
@@ -84,6 +84,24 @@ For the OpenShift 4 Developer Preview we will be sending back these exact attrib
...
   // Subscription has been synced, labelled by name and installed csv
   '{__name__="subscription_sync_total"}',
+  //
+  //OCS metrics to be collected:
+  // ceph_cluster_total_bytes gives the size of ceph cluster in bytes.
+  '{__name__="ceph_cluster_total_bytes"}',
+  // ceph_cluster_total_used_raw_bytes is the amount of ceph cluster storage used in bytes.
+  '{__name__="ceph_cluster_total_used_raw_bytes"}',
+  // ceph_health_status gives the ceph cluster health status
+  '{__name__="ceph_health_status"}',
```
squat: Is this just a 1/0 yes/no? Can this be inferred from the up metric instead of adding a new one?

docs/data-collection.md
```diff
@@ -84,6 +84,24 @@ For the OpenShift 4 Developer Preview we will be sending back these exact attrib
   // subscription_sync_total is the number of times an OLM operator
   // Subscription has been synced, labelled by name and installed csv
   '{__name__="subscription_sync_total"}',
+  //
+  //OCS metrics to be collected:
+  // ceph_cluster_total_bytes gives the size of ceph cluster in bytes.
+  '{__name__="ceph_cluster_total_bytes"}',
```
squat: Naming for summed metrics normally follow the convention of `...units_total`  Is this a metric collected directly from ceph or a recording rule? In the latter case, could we stick to that here as well?

In the former case, what is the cadinality of this metric?

docs/data-collection.md
```diff
@@ -84,6 +84,24 @@ For the OpenShift 4 Developer Preview we will be sending back these exact attrib
...
+  // ceph_cluster_total_used_raw_bytes is the amount of ceph cluster storage used in bytes.
+  '{__name__="ceph_cluster_total_used_raw_bytes"}',
```
squat: Same naming note as above

docs/data-collection.md
```diff
@@ -84,6 +84,24 @@ For the OpenShift 4 Developer Preview we will be sending back these exact attrib
   // subscription_sync_total is the number of times an OLM operator
   // Subscription has been synced, labelled by name and installed csv
   '{__name__="subscription_sync_total"}',
+  //
+  //OCS metrics to be collected:
```
squat: Nit: please insert a space between // and OCS

docs/data-collection.md
anmolsachan: Ack, Done.

docs/data-collection.md
anmolsachan: This has values 0, 1 & 2. This 'ceph_health_status' metric presented is dependent on some internal ceph logic and thus cannot be inferred from 'up' metric.

This has cadinality 1.

docs/data-collection.md
anmolsachan: This is not a recording rule. This is a metric directly collected from ceph. This has cardinality 1.

You can verify this in screenshot 2 of https://github.com/openshift/telemeter/pull/192#issuecomment-514568851

docs/data-collection.md
anmolsachan: @squat This is not a recording rule. This is a metric directly collected from ceph. This has cardinality 1.

You can verify this in screenshot 2 of https://github.com/openshift/telemeter/pull/192#issuecomment-514568851

anmolsachan:
@squat I have updated the PR and addressed the review comments. Please review. Thanks.

PR #193, opened 2019-07-11
Added metric "code:registry_api_request_count:rate:sum" for SLI monitoring

abyrne55:
OpenShift SRE needs to be able to track the rates of API requests to the image registry service for Service Level Agreement compliance. The existing "code:apiserver_request_count:rate:sum" rule doesn't differentiate between the service each API call is going to, so it isn't specific enough for this purpose. To remedy this, I added another recording rule that selects only the API call counters that correspond to the registry service, and then performs a rate-and-sum similar to "code:apiserver_request_count:rate:sum." 

I already added this new recording rule in [this PR for cluster-monitoring-operator](https://github.com/openshift/cluster-monitoring-operator/pull/398). This PR is just for the forwarding rule that tells telemeter to pass the metric along.


abyrne55:
Now that [the PR adding the metric](https://github.com/openshift/cluster-monitoring-operator/pull/398) has been merged, I'm marking this ready for review.

abyrne55:
> looks good, but we will need to rebase on master and regenerate the generated files as there is a conflict

Thanks for taking a look. Rebased and regenerated

smarterclayton:
/hold

we need to resolve policy on these sorts of one off telemetry metrics

PR #194, opened 2019-07-12
metrics: add metrics list for CNV

fromanirh:
kubevirt (https://kubevirt.io) is a kubernetes addon to run virtual
machines on kubernetes clusters.

Signed-off-by: Francesco Romani <fromani@redhat.com>


pkliczewski:
@fromanirh @fabiand @rmohr Shouldn't we use cnv prefix here?

fabiand:
IIUIC this metric name is constraint by what the componet is exposing, in this case we are using a kubevirt mertric - thus we need to use the kubevirt prefix.

To me this is fine in this context at this stage.

docs/data-collection.md
```diff
@@ -81,6 +81,11 @@ For the OpenShift 4 Developer Preview we will be sending back these exact attrib
   // cluster:node_instance_type_count:sum is the number of nodes
   // of each instance type and role.
   '{__name__="cluster:node_instance_type_count:sum"}',
+  //
+  // kubevirt metrics to be collected
+  //
+  // cluster:cnv_vm_status_running:sum is the total number of VM running in the cluster
+  '{__name__="cluster:cnv_vm_status_running:sum"}',
```
smarterclayton: This is probably ok (I assume this is a single series?), are there any error related metrics that track whether there are errors / failures with VMs?

Do you have alerts defined for VM failures on cluster today?

docs/data-collection.md
```diff
@@ -81,6 +81,11 @@ For the OpenShift 4 Developer Preview we will be sending back these exact attrib
   // cluster:node_instance_type_count:sum is the number of nodes
   // of each instance type and role.
   '{__name__="cluster:node_instance_type_count:sum"}',
+  //
+  // kubevirt metrics to be collected
```
smarterclayton: Remove this, I don't think there is anything special about kubevirt.  However, you need to describe in more detail in your description why you want this.  Also make sure it is a complete sentence(s) (punctuation).

docs/data-collection.md
fromanirh: It is meant to be a single series, yes. About the other questions let me forward the question to @fabiand (/cc @msivak)

docs/data-collection.md
fromanirh: Ack, will either drop the comment or add a more informative one.

docs/data-collection.md
smarterclayton: We need to review naming before I approve this (i.e. make sure we're consistent across new metrics types).

rmohr: errors/failures will definitely come in the future. Might make sense to have `running` as a label?

rmohr: @fromanirh is this now reflecting the VMIs or the VMs?

fromanirh: @rmohr makes sense to use the kubevirt naming scheme (VMIs) for consistency, yes. Thanks for the warning about virt-controller, we will see the implementation later.
@smarterclayton Let me provide a bit of context. To provide this metric in cnv we are evaluating two approaches, which will influence the naming we are discussing here.
1: provide low-level metrics in kubevirt, aggregate using a record rule (https://prometheus.io/docs/practices/rules/) provided by a still-to-be-done prometheus mixin. 
Flow: kubevirt metric -> prometheus server -> prometheus record -> telemeter
We will still need to figure out how to deliver such mixin. Is this the preferred way to go? are prometheus mixins supported in openshift?
2: directly expose the final value in kubevirt. This is doable and fits our requirements about VMs count.
Flow: kubevirt metric -> prometheus server -> telemeter

This decision drives the naming scheme we are going to use (we cannot use the "cnv" prefix if we use the second approach), hence the reason I asked. Thanks for your reviews so far.

fromanirh:
I updated the PR addressing the comments and all the internal discussion we had. The prometheus mixin approach (https://github.com/openshift/telemeter/pull/194#discussion_r303788167) is gaining traction because it seems more sustenaible in the long term and because it seems to fit better in the telemeter/prometheus ecosystem. The plan is to deploy such mixin later using one of the kubevirt/cnv specific operators. We are still defining the fine details.

squat:
code-wise this looks fine. Let’s get a final agreement on the naming convention for component-specific metrics and we can merge this (:

fabiand:
lgtm from a naming perspective.

fromanirh:
add to force-push because trivial rebase failed locally

fabiand:
@squat maybe you have a min to approve

smarterclayton:
Name looks ok to me

/approve

pkliczewski:
:tada:

PR #195, opened 2019-07-12
Add OLM metrics to whitelist

ecordell:
Exposes metrics added in: https://github.com/operator-framework/operator-lifecycle-manager/pull/951


docs/data-collection.md
```diff
@@ -74,6 +74,9 @@ For the OpenShift 4 Developer Preview we will be sending back these exact attrib
   '{__name__="cluster:node_instance_type_count:sum"}',
   // cnv:vmi_status_running:count is the total number of VM instances running in the cluster.
   '{__name__="cnv:vmi_status_running:count"}',
+  // subscription_sync_total is the number of times an OLM operator
```
smarterclayton: Approximate cardinality?

smarterclayton: Naming wise this should probably `operator_subscription_sync_total` - subscription is kind of vague, although on an Openshift cluster OLM definitely can make the case for being generic.

ecordell: It's O(installed operators), my guess is <50 per cluster

Do you want me to change the metric name or leave it as is?

smarterclayton: metric name is fine, we can change later.

PR #198, opened 2019-07-16
metrics: Remove a metric that is no longer reported

smarterclayton:
cluster_version_payload_errors was removed prior to 4.1 GA when we
switched to an "always retry" mode for payload.


squat:
I just rebased PR (:

squat:
rebased again now that #197 landed

PR #204, opened 2019-07-24
Add Route annotation to ServiceAccount for thanos-querier

metalmatze:
We need to update the annotations on the prometheus-telemeter to use its ServiceAccount for the Thanos querier too.

/cc @aditya-konarde @brancz @squat

### DIFF
--- a/jsonnet/telemeter/prometheus/kubernetes.libsonnet
+++ b/jsonnet/telemeter/prometheus/kubernetes.libsonnet
@@ -41,6 +41,7 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
       serviceAccount.new('prometheus-' + $._config.prometheus.name) +
       serviceAccount.mixin.metadata.withNamespace($._config.namespace) +
       serviceAccount.mixin.metadata.withAnnotations({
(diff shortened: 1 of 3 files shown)


jsonnet/telemeter/prometheus/kubernetes.libsonnet
```diff
@@ -41,6 +41,7 @@ local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
       serviceAccount.new('prometheus-' + $._config.prometheus.name) +
       serviceAccount.mixin.metadata.withNamespace($._config.namespace) +
       serviceAccount.mixin.metadata.withAnnotations({
+        'serviceaccounts.openshift.io/oauth-redirectreference.observatorium-thanos-querier':'{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"observatorium-thanos-querier"}}',
```
squat: Can we add a TODO here to eliminate this once we have a real clusterrole?

metalmatze: done! :+1:

PR #208, opened 2019-07-31
OBS-17: implement client credentials grant flow

s-urbaniak:
This change configures telemeter server to use the client credentials grant flow if the authorize username is unset.

This should make the transition to sso.redhat.com (using client credentials grant) from developers.redhat.com (using resource owner password credentials grant) very easy, by simply setting RHD_USERNAME to empty string `""`, and just reconfiguring RHD_CLIENT_ID, RHD_PASSWORD, and RHD_URL.

Fixes OBS-17

/cc @squat @jmelis @LiliC @ant31


squat:
nice :) /lgtm

squat:
/lgtm
nice (:

PR #209, opened 2019-08-02
jsonnet: Include workload:*:sum metrics in reporting

smarterclayton:
Infrastructure usage is now measured as cluster usage - workload usage
to better capture host level measurements. This changes telemetry to
report that additional value for comparison.

Depends on https://github.com/openshift/cluster-monitoring-operator/pull/425 which is merged


LiliC:
/cherrypick release-4.1

See https://github.com/openshift/cluster-monitoring-operator/pull/432#issuecomment-520504198

PR #210, opened 2019-08-02
Bug 1737164: Add OLM metrics to whitelist

ecordell:
cherry pick of #195


crawford:
The release architects have agreed that there is no need for a valid Bugzilla for this particular backport due to it being purely telemetry-focused and because QE isn't able to verify.

PR #212, opened 2019-08-05
metrics: gather cluster_installer series

crawford:
This series will allow us to determine whether a cluster was created
using IPI or UPI. Once Hive and CI are updated, we'll also be able to
determine if a cluster was created by CI or if it was created by Hive.

The series captures the install type, the version of the tool, and its
invoker (in addition to the standard labels):

    cluster_installer{
        type="openshift-install",
        version="unreleased-master-1209-gfd08f44181f2111",
        invoker="user",
    } 0 1562168623759

This is a backport of https://github.com/openshift/telemeter/pull/189.


crawford:
The release architects have agreed that there is no need for a valid Bugzilla for this particular backport due to it being purely telemetry-focused and because QE isn't able to verify.

PR #213, opened 2019-08-06
pkg/store/forward: defensively override timestamps

squat:
This commit ensures that time series forwarded by telemeter server to a
remote-write endpoint will defensively have their timestamps
overwritten, ensuring that they are always increasing monotonically.

cc @metalmatze @brancz

### DIFF
--- a/pkg/store/forward/forward.go
+++ b/pkg/store/forward/forward.go
@@ -18,7 +18,9 @@ import (
 	"github.com/openshift/telemeter/pkg/store"
 )
 
-const metricName = "__name__"
+const (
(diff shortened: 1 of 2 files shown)


brancz:
Thinking about this some more ... I think we should probably only overwrite time stamps when absolutely necessary and have a metric for when we do so.

squat:
Yes we can definitely do that. I was thinking about only overwriting when
the drift is negative (timestamp is in the future) and increment a
“timestamp_overwritten” counter

brancz:
Sounds good to me.

PR #214, opened 2019-08-07
Log the mean drift from current timestamp when forward fails

metalmatze:
For every failed request we will get the mean drift from the current timestamp. This should be helpful to find clusters that are way off.

/cc @squat @kakkoyun @brancz

### DIFF
--- a/pkg/store/forward/forward.go
+++ b/pkg/store/forward/forward.go
@@ -108,13 +108,16 @@ func (s *Store) WriteMetrics(ctx context.Context, p *store.PartitionedMetrics) e
 				WithLabelValues(fmt.Sprintf("%d", resp.StatusCode)).
 				Observe(time.Since(begin).Seconds())
 
-			if resp.StatusCode/100 != 2 {
-				meanDrift := timeseriesMeanDrift(timeseries, time.Now().Unix())
-				return fmt.Errorf("response status code is %s - mean drift from now for clusters %s is: %.3fs",
-					resp.Status,
+			meanDrift := timeseriesMeanDrift(timeseries, time.Now().Unix())
+			if meanDrift > 10 {
+				log.Printf("mean drift from now for clusters %s is: %.3fs",
 					p.PartitionKey,
 					meanDrift,
 				)
+			}
+
+			if resp.StatusCode/100 != 2 {
+				return fmt.Errorf("response status code is %s", resp.Status)
 			}
 
 			s := 0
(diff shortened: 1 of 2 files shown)


pkg/store/forward/forward.go
```diff
@@ -109,7 +109,12 @@ func (s *Store) WriteMetrics(ctx context.Context, p *store.PartitionedMetrics) e
 				Observe(time.Since(begin).Seconds())
 
 			if resp.StatusCode/100 != 2 {
-				return fmt.Errorf("response was not 200 OK, but %s", resp.Status)
+				mean := timeseriesMeanDrift(timeseries, time.Now().Unix()*1000)
+				return fmt.Errorf("response status code is %s - mean drift from now for clusters %S is: %.3fms",
+					resp.Status,
+					p.PartitionKey,
+					float64(time.Now().Second())-mean,
```
squat: The mean drift will be something like `200ms`, we dont need to subtract that again from `now`, right?

pkg/store/forward/forward.go
```diff
@@ -109,7 +109,12 @@ func (s *Store) WriteMetrics(ctx context.Context, p *store.PartitionedMetrics) e
 				Observe(time.Since(begin).Seconds())
 
 			if resp.StatusCode/100 != 2 {
-				return fmt.Errorf("response was not 200 OK, but %s", resp.Status)
+				mean := timeseriesMeanDrift(timeseries, time.Now().Unix()*1000)
+				return fmt.Errorf("response status code is %s - mean drift from now for clusters %S is: %.3fms",
```
squat: `%S` should be `%s`

pkg/store/forward/forward.go
```diff
@@ -109,7 +109,12 @@ func (s *Store) WriteMetrics(ctx context.Context, p *store.PartitionedMetrics) e
 				Observe(time.Since(begin).Seconds())
 
 			if resp.StatusCode/100 != 2 {
-				return fmt.Errorf("response was not 200 OK, but %s", resp.Status)
+				mean := timeseriesMeanDrift(timeseries, time.Now().Unix()*1000)
```
squat: looking at this again, the second param for `timeseriesMeanDrift` is seconds, not milliseconds, so we don't need the `*1000`

pkg/store/forward/forward.go
squat: also, looking at it again, `timeseriesMeanDrift`, returns a value in seconds, not millis, so it should be `%.3fs`.

pkg/store/forward/forward.go
metalmatze: Exactly. I already calculate the mean of the drift (which is now - timestamp) per timestamp.

PR #216, opened 2019-08-07
Makefile: bump benchmark goal

squat:
ran a new round of benchmarks and found we can currently handle 9000 clusters

cc @metalmatze @kakkoyun


squat:
Hmm the stack isn’t failing to handle the load, instead the benchmark tool is using so much memory that it gets killed. Let’s see if we can find a workaround. Otherwise we’ll have to lower the worker count

PR #217, opened 2019-08-08
Log mean drift time for cluster more than 10s apart

metalmatze:
We want to also log clusters that ingest data being in the future with a mean more than 10s.

/cc @squat


squat:
thank you, robot

metalmatze:
:dog: 

/woof

PR #218, opened 2019-08-09
 pkg/store/forward: only overwrite future timestamps

squat:
Only overwrite timestamps that are in the future.
Also expose a metric to track how many timestamps are being overwritten.

cc @brancz @metalmatze


squat:
xref: https://github.com/openshift/telemeter/pull/213#issuecomment-519649600

PR #220, opened 2019-08-09
jsonnet/telemeter: remove auth username

squat:
remove the username environment variable and arg from the telemeter-server statefulset so we can use the client-credentials grant flow.

cc @jmelis @metalmatze

### DIFF
--- a/jsonnet/telemeter/server/kubernetes.libsonnet
+++ b/jsonnet/telemeter/server/kubernetes.libsonnet
@@ -16,7 +16,6 @@ local clusterPort = 8082;
       authorizeURL: 'https://api.openshift.com/api/accounts_mgmt/v1/cluster_registrations',
       replicas: 10,
       rhdURL: '',
-      rhdUsername: '',
       rhdPassword: '',
       rhdClientID: '',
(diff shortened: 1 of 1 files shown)


jsonnet/telemeter/server/kubernetes.libsonnet
```diff
@@ -126,7 +123,6 @@ local clusterPort = 8082;
 
       secret.new(secretName, {
         'rhd.url': std.base64($._config.telemeterServer.rhdURL),
-        'rhd.username': std.base64($._config.telemeterServer.rhdUsername),
         'rhd.password': std.base64($._config.telemeterServer.rhdPassword),
```
brancz: is "password" still appropriate here? would "secret" not be more appropriate?

squat: yes, we can make that change. then the args will have `--password=$SECRET` because https://github.com/openshift/telemeter/pull/208 recycles the same args

brancz: Ack, let's do all of that in one go

PR #223, opened 2019-08-20
Add a receive handler to Telemeter for remote-write

metalmatze:
We want to add a remote-write receive handler to Telemeter.
For this we need to add a new authorizer, as we are unable to use JWTs.

Works with a mock authorization-server.
Here's a Prometheus config to test with:

```yaml
global:
  scrape_interval:     15s
  evaluation_interval: 15s

scrape_configs:
  - job_name: 'prometheus'
    static_configs:
    - targets: ['localhost:9090']
  - job_name: 'telemeter'
    static_configs:
    - targets: ['localhost:9004']

remote_write:
  - url: "http://localhost:9003/metrics/v1/receive"
    bearer_token: 'eyJjbHVzdGVyIjogImU1MGFiNGIwLTgwZjgtNDU3OC05NWVmLWQ0NzI5Njg5NzZhZSIsICJ0b2tlbiI6ICJhIn0K'
```

TODOs:
* [ ] Add some basic metrics (compare with forward store)

/cc @squat @brancz @kakkoyun

### DIFF
--- a/cmd/telemeter-server/main.go
+++ b/cmd/telemeter-server/main.go
@@ -116,6 +116,7 @@ func main() {
 	cmd.Flags().StringVar(&opt.OIDCIssuer, "oidc-issuer", opt.OIDCIssuer, "The OIDC issuer URL, see https://openid.net/specs/openid-connect-discovery-1_0.html#IssuerDiscovery.")
 	cmd.Flags().StringVar(&opt.ClientSecret, "client-secret", opt.ClientSecret, "The OIDC client secret, see https://tools.ietf.org/html/rfc6749#section-2.3.")
 	cmd.Flags().StringVar(&opt.ClientID, "client-id", opt.ClientID, "The OIDC client ID, see https://tools.ietf.org/html/rfc6749#section-2.3.")
+	cmd.Flags().StringVar(&opt.TenantKey, "tenant-key", opt.TenantKey, "The JSON key in the bearer token whose value to use as the tenant ID.")
 
 	cmd.Flags().DurationVar(&opt.Ratelimit, "ratelimit", opt.Ratelimit, "The rate limit of metric uploads per cluster ID. Uploads happening more often than this limit will be rejected.")
 	cmd.Flags().DurationVar(&opt.TTL, "ttl", opt.TTL, "The TTL for metrics to be held in memory.")
@@ -155,6 +156,7 @@ type Options struct {
 	OIDCIssuer   string
 	ClientID     string
 	ClientSecret string
+	TenantKey    string
 
 	PartitionKey      string
 	LabelFlag         []string
@@ -356,9 +358,7 @@ func (o *Options) Run() error {
 	secret := h.Sum(nil)[:32]
 
 	external := http.NewServeMux()
-	externalProtected := http.NewServeMux()
 	internal := http.NewServeMux()
-	internalProtected := http.NewServeMux()
 
 	internalPaths := []string{"/", "/federate", "/metrics", "/debug/pprof", "/healthz", "/healthz/ready"}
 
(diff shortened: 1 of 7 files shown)


pkg/receive/handler.go
```diff
@@ -0,0 +1,114 @@
...
+
+	if resp.StatusCode/100 != 2 {
+		log.Printf("response status code is %s\n", resp.Status)
+		http.Error(w, "upstream response status is not 200 OK", http.StatusBadGateway)
+		return
+	}
+}
+
+type AuthorizationPayload struct {
+	Cluster string `json:"cluster"`
```
brancz: This'll work for now, but doesn't generally work. We need to rethink this a bit. Not a blocker on this though.

cmd/telemeter-server/main.go
```diff
@@ -508,6 +510,13 @@ func (o *Options) Run() error {
 	}))
 	telemeter_http.HealthRoutes(external)
 	external.Handle("/authorize", telemeter_http.NewInstrumentedHandler("authorize", auth))
+	external.Handle("/receive",
```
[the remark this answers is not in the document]
squat: Yes I was thinking the same. That keeps our options open as it is both API group and version.

pkg/receive/handler.go
squat: Ack, this is 100% specific to Tollbooth. We’ll need to figure out a way to generalize this and maintain compat with Tollbooth auth

squat: I just realized that there is a decent way we can keep this generic and not tightly coupled to tollbooth. tollbooth expects all authorization requests to be POST requests with the following body:
```json
{"cluster_id":"<cluster_id>","authorization_token":"<pull_secret_auth_token>"}
```
We can ensure that the bearer token that prometheus is configured to send matches that format and then we POST that token to the specified URL. This way we don't have to do any custom marshalling/unmarshalling. The only variables then are the authorization URL and the token contents.

The one thing that is funny is that we are taking the bearer token header and using that as the body to of a POST request, but I still think that's better and more generic than what we have today.

cmd/telemeter-server/main.go
```diff
@@ -472,9 +473,10 @@ func (o *Options) Run() error {
 	transforms.With(metricfamily.NewElide(o.ElideLabels...))
 
 	server := httpserver.New(store, validator, transforms, o.TTL)
+	receiver := receive.NewHandler(o.ForwardURL)
 
 	internalPathJSON, _ := json.MarshalIndent(Paths{Paths: internalPaths}, "", "  ")
-	externalPathJSON, _ := json.MarshalIndent(Paths{Paths: []string{"/", "/authorize", "/upload", "/healthz", "/healthz/ready"}}, "", "  ")
+	externalPathJSON, _ := json.MarshalIndent(Paths{Paths: []string{"/", "/authorize", "/upload", "/healthz", "/healthz/ready", "receive"}}, "", "  ")
```
squat: for consistency we need a preceding `/` here

pkg/authorize/handler.go
```diff
@@ -31,3 +37,106 @@ func NewAuthorizeClientHandler(authorizer ClientAuthorizer, next http.Handler) h
...
+	defer func() {
+		// read the body to keep the upstream connection open
+		if res.Body != nil {
+			if _, err := io.Copy(ioutil.Discard, res.Body); err != nil {
+				log.Printf("error copying body: %v", err)
+			}
+			res.Body.Close()
+		}
+	}()
+	body, err := ioutil.ReadAll(io.LimitReader(res.Body, 32*1024))
```
metalmatze: Let's make 32*1024 a constant. :)

PR #225, opened 2019-08-22
Add metric "knative_serving_health" to monitor health status of knative serving

savitaashture:
knative serving operator expose metric called `knative_serving_health` which shows health status of installed knative serving

image

### DIFF
--- a/docs/data-collection.md
+++ b/docs/data-collection.md
@@ -127,6 +127,8 @@ For the OpenShift 4 Developer Preview we will be sending back these exact attrib
   '{__name__="noobaa_total_usage"}',
   // console_url is the url of the console running on the cluster.
   '{__name__="console_url"}',
+  // knative_serving_health gives the health status of installed knative serving.
+  '{__name__="knative_serving_health"}',
 ]
 ```
 
--- a/jsonnet/telemeter/metrics.jsonnet
+++ b/jsonnet/telemeter/metrics.jsonnet
@@ -119,4 +119,6 @@
   '{__name__="noobaa_total_usage"}',
   // console_url is the url of the console running on the cluster.
   '{__name__="console_url"}',
+  // knative_serving_health gives the health status of knative serving.
+  '{__name__="knative_serving_health"}',
 ]
--- a/manifests/client/deployment.yaml
+++ b/manifests/client/deployment.yaml
@@ -66,6 +66,7 @@ spec:
         - --match={__name__="noobaa_accounts_num"}
         - --match={__name__="noobaa_total_usage"}
         - --match={__name__="console_url"}
+        - --match={__name__="knative_serving_health"}
         env:
         - name: ANONYMIZE_LABELS
           value: ""


smarterclayton:
/hold

I would expect this to be reported by OLM, not by a direct metric.

bbrowning:
@smarterclayton I had the exact same thought initially, but from OLM's point of view our operator is "OpenShift Serverless". When a user installs the OpenShift Serverless operator, they will then be able to choose whether to install Knative Serving, Knative Eventing, etc control planes by creating the appropriate custom resource. So, OLM will tell us what versions of this OpenShift Serverless operator are installed. But what we won't be able to see is how many people using that operator chose to install Serving, Eventing, or both.

smarterclayton:
If we want to track that, I think we’re going to want a more specific
metric exposed by the operator - ideally you surface a consistent metric
(with one or more series) that contains labels and captures the core usage
you need.  For instance, a better metric is the number of knative services
in use, or the number of event sources/sinks.  Version should be determined
by your operator and will be a standardized by OLM - having a single
business metric that represents use will be far more valuable for tracking
adoption.

On Aug 22, 2019, at 5:03 PM, Ben Browning <notifications@github.com> wrote:

@smarterclayton <https://github.com/smarterclayton> I had the exact same
thought initially, but from OLM's point of view our operator is "OpenShift
Serverless". When a user installs the OpenShift Serverless operator, they
will then be able to choose whether to install Knative Serving, Knative
Eventing, etc control planes by creating the appropriate custom resource.
So, OLM will tell us what versions of this OpenShift Serverless operator
are installed. But what we won't be able to see is how many people using
that operator chose to install Serving, Eventing, or both.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/telemeter/pull/225?email_source=notifications&email_token=AAI37J23TMKLHQEJR3GB6JTQF2TDJA5CNFSM4IOTI3LKYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOD45MITY#issuecomment-523945039>,
or mute the thread

smarterclayton:
This still looks like health, and OLM is still planning on tracking this, right?

markusthoemmes:
The discussion above revolved around version though. I don't think OLM could surface meaningful healthiness metrics without actually known the thing itself? Or am I missing something?

markusthoemmes:
For posterity (from our Slack DM):
Metrics per your proposal (# of services etc) might and will follow. This health metric is mostly to track if a rollout is going wrong and if our installation experience is good enough for people to be successful at installing Knative Serving as part of the Serverless Operator.

As such it's not something OLM directly knows about. It's more akin to an Etcd instance as it would be deployed by the Etcd Operator.

bbrowning:
Should we close this PR and revisit when we pick up work to integrate with telemetry again?

squat:
@bbrowning yes sorry this went stale. The canonical location for registering new telemeter metrics is now https://github.com/openshift/cluster-monitoring-operator/blob/master/manifests/0000_50_cluster_monitoring_operator_04-config.yaml#L4

bbrowning:
@squat No worries - the staleness was mostly due to needing to revisit integration on our end. Let's see if Prow lets me close this one.

/close

PR #228, opened 2019-08-28
Use scoped Prometheus Registry instead of global

metalmatze:
/cc @squat

### DIFF
--- a/pkg/http/server.go
+++ b/pkg/http/server.go
@@ -7,42 +7,35 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
-var (
-	requestDuration = prometheus.NewHistogramVec(
+func NewInstrumentedHandler(reg *prometheus.Registry, handlerName string, next http.Handler) http.Handler {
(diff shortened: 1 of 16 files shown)


pkg/http/server.go
```diff
@@ -3,46 +3,58 @@ package http
...
 
-func NewInstrumentedHandler(handlerName string, next http.Handler) http.Handler {
-	return promhttp.InstrumentHandlerDuration(
-		requestDuration.MustCurryWith(prometheus.Labels{"handler": handlerName}),
-		promhttp.InstrumentHandlerRequestSize(
-			requestSize.MustCurryWith(prometheus.Labels{"handler": handlerName}),
-			promhttp.InstrumentHandlerCounter(
-				requestsTotal.MustCurryWith(prometheus.Labels{"handler": handlerName}),
+func (i InstrumentHandler) Handle(handlerName string, next http.Handler) http.Handler {
+	labels := prometheus.Labels{"handler": handlerName}
```
squat: :sunglasses:

metalmatze:
We actually don't care too much about this compared to the time we would need to improve this...

PR #229, opened 2019-08-28
Bug 1746686: jsonnet: fix client reload port

squat:
The configmap reload for telemeter client was pointing to an unused port,
meaning that it would never reload anything.

cc @s-urbaniak @metalmatze @kakkoyun

### DIFF
--- a/jsonnet/telemeter/client/kubernetes.libsonnet
+++ b/jsonnet/telemeter/client/kubernetes.libsonnet
@@ -9,7 +9,7 @@ local servingCertsCABundle = 'serving-certs-ca-bundle';
 local servingCertsCABundleFileName = 'service-ca.crt';
 local servingCertsCABundleMountPath = '/etc/%s' % servingCertsCABundle;
 local fromTokenFile = '/var/run/secrets/kubernetes.io/serviceaccount/token';
-local metricsPort = 8080;
+local insecurePort = 8080;
 local securePort = 8443;
 
 {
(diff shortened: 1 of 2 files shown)


s-urbaniak:
I rechecked and indeed we are not having the correct port specified in the reload container:
```
$ oc -n openshift-monitoring rsh -c reload telemeter-client-688d8c599f-f6wct curl localhost:9000
curl: (7) Failed connect to localhost:9000; Connection refused
```

I checked why we have not yet fallen into that regression and the reason is probably that the serving cert certificates are valid for one year:
```
$ openssl crl2pkcs7 -nocrl -certfile <(oc -n openshift-monitoring rsh -c reload telemeter-client-688d8c599f-f6wct cat /etc/serving-certs-ca-bundle/service-ca.crt) | openssl pkcs7 -print_certs -noout -text | grep -e 'Not Before:\|Not After :\|Issuer:\|Subject:'
        Issuer: CN=openshift-service-serving-signer@1566891773
            Not Before: Aug 27 07:42:53 2019 GMT
            Not After : Aug 26 07:42:54 2020 GMT
        Subject: CN=openshift-service-serving-signer@1566891773
```

Is there a way we could add an integration test here which asserts the correct port assignment between the reloader and telemeter client?

I'll open a 4.2 and 4.1 BZ so we can backport.

squat:
@s-urbaniak thanks for the investigation. I could have sworn that the certs expired sooner. So the certs are rotated every few hours but the CA that signs them is valid for a year.

Now I really want to make sure to verify that this reloading configuration _actually_ works e2e. Let’s discuss further oob.

s-urbaniak:
@s-urbaniak indeed, maybe something changed with the rotation. I rechecked the api server aggregation certs which used to be also very aggressive (4 hours) but are now valid for months/years.

PR #230, opened 2019-08-28
jsonnet: deploy telemeter benchmark

squat:
As discussed out of band, we want to generate production-like load against our integration environments. This PR creates configuration to deploy the telemeter benchmark tool as a service.
The only thing that needs to be adjusted in order to work correctly is to define a valid token in the telemeter-benchmark secret.
This is running on our long-lived test cluster against staging and working :100: 

cc @brancz @metalmatze @kakkoyun @aditya-konarde @kbsingh @cben


jsonnet/telemeter/benchmark/kubernetes.libsonnet
```diff
@@ -9,12 +9,21 @@ local authorizePort = 8083;
...
 local tokensFileName = 'tokens.json';
+local timeseriesFileName = 'timeseries.txt';
+local configmapMountPath = '/etc/config';
 
 {
   _config+:: {
     namespace: 'telemeter-benchmark',
 
+    telemeterBenchmark+:: {
+      to: 'https://infogw.api.openshift.com',
```
cben: Will this add data to production telemeter?
I was expecting you'd do infogw.api.stage.
(Is there a third one for integration env?)

cben:
cc @aditya-konarde FYI.

cben:
I guess what I don't know is which downstream components other than telemeter itself will this affect.
Will this increase authorization requests to account-manager?  
Will this add fake data visible to queries & federation?

squat:
@cben to answer your questions:

1. That url is just a placeholder. The actual deployment manifest used by sass-herder will be a resource list template with that value as a variable. I’ll make that in a follow up PR so as to not make this or too huge.
2. These are regular telemeter requests. From the point of view of Telemeter they are indistinguishable from "real" requests. This means they will affect every downstream component.
3. See 2. They are real requests so the data will will be queryable, and will end up in telemeter-prometheus, the Observatorium s3 bucket, and telemeter-cache.

PR #231, opened 2019-09-04
Bump prometheus-operator to 0.32

LiliC:
We need this to bump prometheus-operator in CMO, currently, it fails as we have a mismatch of versions.

cc @s-urbaniak


s-urbaniak:
:+1:

PR #235, opened 2019-09-13
 Bug 1749451: Capture sum of cpu capacity cores by OS (RHCOS or RHEL), arch, and role for the nodes. 

chambridge:
Add metric gathering for node_role_os_version_machine:cpu_capacity_cores:sum.

The following metric was added to `cluster-monitoring-operator` here: https://github.com/openshift/cluster-monitoring-operator/pull/466

Intended to be utilized for subscription reconciliation when cluster has mixed set of nodes (bring your own worker) and node roles.


LiliC:
/hold

We want to update the saas-telemeter at the same time

PR #238, opened 2019-09-19
Overwrite all timestamps for Telemeter v1 & v2

metalmatze:
In order to be always consistent with data between Telemeter v1 and v2, we want to explicitly always overwrite the timestamp to the current time.

In the past, Telemeter v1 simply dropped all timestamps, leaving Telemeter Prometheus making that decision (which used its current timestamp upon ingestion). Now in the future, we will consistently forward the current timestamp to both v1 and v2.

/cc @aditya-konarde @brancz @squat @s-urbaniak @kakkoyun @bwplotka


brancz:
This generally lgtm but would like @s-urbaniak to have a look as well.

pkg/metricfamily/override.go
```diff
@@ -0,0 +1,48 @@
...
+		Name: "telemeter_overwritten_timestamps_total",
+		Help: "Number of timestamps that were in the past, present or future",
+	}, []string{"tense"})
+)
+
+func init() {
+	prometheus.MustRegister(overwrittenMetrics)
+}
+
+func OverwriteTimestamps(now func() time.Time) TransformerFunc {
```
s-urbaniak: `now func() time.Time` is very cool for testing :+1: do you mind to add one, is it worth?

s-urbaniak: maybe let's add a little godoc for the future on why and what this transformer is doing.

s-urbaniak:
@metalmatze it might be, that we need to adjust some invariants in the unit tests :thinking: those (i think) still assume timestamp dropping.

pkg/metricfamily/override.go
metalmatze: Haha, actually went ahead and did exactly that without reading this comment. :smile: :+1:

PR #241, opened 2019-09-24
WIP: Add remote write config for subscription_labels

kakkoyun:
This PR adds remote write config to Prometheus, to send `subscription labels` to Telemeter V2


kakkoyun:
cc @metalmatze @aditya-konarde

brancz:
Adding this to telemeter Prometheus may not be a great idea as remote write adds a non trivial memory overheard for metadata management. I think this might be better as a separate Prometheus instance.

PR #245, opened 2019-10-08
Bug 1746711: fix client reload port

LiliC:
The configmap reload for telemeter client was pointing to an unused port,
meaning that it would never reload anything.

Backport to the 4.1 release of https://github.com/openshift/telemeter/pull/229/files

cc @s-urbaniak @squat

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 .PHONY: all build image check test-generate test-integration test-benchmark vendor dependencies manifests
 
+SHELL=/bin/bash
 BIN=bin
 GOLANGCI_LINT_BIN=$(BIN)/golangci-lint
 EMBEDMD_BIN=$(GOPATH)/bin/embedmd
@@ -38,7 +39,7 @@ docs/telemeter_query: $(JSONNET_SRC)
 	echo "$$query" > $@
 
 test-generate:
-	make --always-make && git diff --exit-code
+	make manifests && git diff --exit-code
 
 lint: $(GOLANGCI_LINT_BIN)
 	# megacheck fails to respect build flags, causing compilation failure during linting.
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ docs/telemeter_query: $(JSONNET_SRC)
 	echo "$$query" > $@
 
 test-generate:
-	make manifests && git diff --exit-code
+	make manifests --always-make && git diff --exit-code
 
 lint: $(GOLANGCI_LINT_BIN)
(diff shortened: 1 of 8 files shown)


s-urbaniak:
but it seems we have an issue with `jb` :thinking: 
```
cd jsonnet && jb install
Initialized empty Git repository in /go/src/github.com/openshift/telemeter/jsonnet/vendor/.tmp/jsonnetpkg-telemeter-e580d1f71da91e809c0f52dc6272536cf08605a9599261776/.git/
fatal: '../' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: '../' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
jb: error: failed to install: failed to install package: exit status 128
```

LiliC:
/hold cancel

😅 that was a ride to get everything right

LiliC:
CI should be fixed when https://github.com/openshift/release/pull/5336 is merged.

Makefile
```diff
@@ -38,7 +39,7 @@ docs/telemeter_query: $(JSONNET_SRC)
 	echo "$$query" > $@
 
 test-generate:
-	make --always-make && git diff --exit-code
+	make manifests && git diff --exit-code
```
squat: We *have* to have `--always-make` here, otherwise we can't guarantee that the manifests will be re-generated for the test, in which case the test won't actually compare anything new.

Makefile
```diff
@@ -88,17 +89,17 @@ $(JSONNET_VENDOR): jsonnet/jsonnetfile.json $(JB_BIN)
...
 
 $(JB_BIN):
-	go get -u github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb
+	GO111MODULE=off go get -u github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb
 
 $(JSONNET_BIN):
-	go get -u -d github.com/google/go-jsonnet/cmd/jsonnet
-	cd $(GOPATH)/src/github.com/google/go-jsonnet && git checkout v0.12.1 && git submodule update && go install -a ./jsonnet
+	GO111MODULE=off go get -u -d github.com/google/go-jsonnet/cmd/jsonnet
+	cd $(GOPATH)/src/github.com/google/go-jsonnet
```
squat: This `cd` should be removed now that the `git checkout` is gone

LiliC:
@squat PTAL thanks!

PR #247, opened 2019-10-10
list Roles before RoleBindings

maorfr:
As Roles are listed after RoleBindings, first deployment will always fail:

https://ci-int-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/telemeter/job/openshift-telemeter-gh-build-master/63/console


squat:
Hi @maorfr, these files are all automatically generated by jsonnet, so reordering or renaming them by hand will not work. Please make the changes in the jsonnet file. If you need help, we hand happily take a look. As a side note, applying these manifests by hand using kubectl or oc does not result in any ordering issues for me. Is this a problem with the saas-herder tooling?

maorfr:
sorry for the delay.
thanks for your reply @squat!

i understand, and i will try to implement accordingly.
its only a problem the first time we apply, we've seen this fail once or twice so far.

maorfr:
closing this as we've decided to disable that check on the deployment pipeline.

PR #248, opened 2019-10-10
Bug 1774623: Add marketplace metrics for default OperatorSources

awgreene:
Exposes metrics added in: https://github.com/operator-framework/operator-marketplace/pull/246


brancz:
These seem like pretty large cardinality and each team only gets 10 time-series. Probably the first metric already exceeds this on an order of magnitude. We’ll need to first create aggregates of this and then send only the aggregates.

awgreene:
https://github.com/openshift/telemeter/pull/248#issuecomment-540921544

Thank you for your review @brancz! In regards to your feedback:

> These seem like pretty large cardinality and each team only gets 10 time-series.

OLM/Marketplace was granted permission to create more than 10 time-series, I can forward you an email with the details if you would like.

> Probably the first metric already exceeds this on an order of magnitude.  We’ll need to first create aggregates of this and then send only the aggregates.

This is a great point. I will remove the `app_registry_request_duration_seconds` metric - it will be useful on cluster and does not need to be exported to Telemeter.

I will aggregate the app_registry_request_total time-series as you suggested. I would like to record:
- The http status codes based on their first digit  (100,200,300,400,500)
- The quay namespace(`community-operators`, `certified-operators`, `redhat-operators`)

This would lead to a potential maximum of 5*3 = 15 time-series.

brancz:
I think that would work. Could you ping us on the PR where you introduce that recording rule? :)

awgreene:
@brancz 

> I think that would work. Could you ping us on the PR where you introduce that recording rule? :)

Wonderful, will do!

awgreene:
@brancz A [PR](https://github.com/operator-framework/operator-marketplace/pull/259/files) against Marketplace has been opened to introduce the proposed changes.

awgreene:
> It seems like there is a mismatch between the generated files and those specified in the `jsonnet/telemeter/metrics.jsonnet `, can you rerun `make generate-in-docker` again, thanks
> 
> https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_telemeter/248/pull-ci-openshift-telemeter-master-generate/165#1:build-log.txt%3A83

Hello @LiliC, it appears that the container is failing:
```
$ make generate-in-docker
find: ‘./benchmark’: No such file or directory
docker run --rm -u="1000:1000" -v "/home/agreene/.cache/go-build:/.cache/go-build" -v "/home/agreene/go/src/github.com/openshift/telemeter:/go/src/github.com/openshift/telemeter:Z" -w "/go/src/github.com/openshift/telemeter" -e GO111MODULE=on quay.io/coreos/jsonnet-ci make  generate
Unable to find image 'quay.io/coreos/jsonnet-ci:latest' locally
latest: Pulling from coreos/jsonnet-ci
c7b7d16361e0: Pull complete 
b7a128769df1: Pull complete 
1128949d0793: Pull complete 
667692510b70: Pull complete 
c70d80036479: Pull complete 
07b6f8daecbf: Pull complete 
f41babb1524b: Pull complete 
7779ba8293d3: Pull complete 
bcff02fef15d: Pull complete 
d80978f814fd: Pull complete 
8c8dd5e167a3: Pull complete 
Digest: sha256:324dd79f91898680333556caf846f945c349763c9ca8f010335dcc88742e200c
Status: Downloaded newer image for quay.io/coreos/jsonnet-ci:latest
find: './benchmark': No such file or directory
rm -rf manifests
mkdir -p manifests/{benchmark,client,server,prometheus}
/go/bin/jsonnet jsonnet/benchmark.jsonnet -J jsonnet/vendor -m manifests/benchmark
RUNTIME ERROR: couldn't open import "prometheus-operator/prometheus-operator.libsonnet": no match locally or in the Jsonnet library paths
	jsonnet/telemeter/benchmark.libsonnet:4:12-70	thunk <b> from <$>
	jsonnet/telemeter/benchmark.libsonnet:34:22-23	
	jsonnet/benchmark.jsonnet:3:84-101	thunk from <$>
	<std>:1268:24-25	thunk from <function <anonymous>>
	<std>:1268:5-33	function <anonymous>
	jsonnet/benchmark.jsonnet:3:67-102	$
		
		
	During evaluation	

make: *** [Makefile:103: manifests] Error 1
make: *** [Makefile:81: generate-in-docker] Error 2
```

Do you have any suggestions on how to get past this step?

LiliC:
@awgreene sorry about not seeing this before, travelling for conferences. Seems like the problem is solved now. What did you end up doing, so we can be clearer in our docs. :)

awgreene:
@LiliC When I had originally opened this PR the docs did not exist, so reading the docs gave me the background knowledge I needed to generate the files.

awgreene:
@LiliC are there any other required changes?

LiliC:
No should be good to go, letting @brancz have a look. Not sure we can merge this right now in 4.3, unless you have a bugzilla as its code freeze already.

awgreene:
@LiliC I have created a Bugzilla and added attached it to the PR.

brancz:
If I recall correctly, there are already alerts that use these metrics to alert on. The alerts are automatically picked up, so this seems duplicate.

awgreene:
>If I recall correctly, there are already alerts that use these metrics to alert on. The alerts are automatically picked up, so this seems duplicate.

We had originally wanted the buckets to better understand the outages, but after considering your point we have decided that the alerts are sufficient.

PR #259, opened 2019-10-31
Metrics to collect the usage of multus cni plugin

aneeshkp:
This PR is for collecting usage of multus CNI plugin labeled by networks.
These are the related PR's to expose metrics to Prometheus from the instance
https://github.com/openshift/cluster-network-operator/pull/377
(Yet to be merged)


brancz:
If I'm understanding correctly then the cardinality of this should be ok, but just to make sure can you share a screenshot of you querying these metrics on a live cluster?

aneeshkp:
@brancz  network_attachment_definition_enabled_instance_up will have only two varieties  based on the label `network_attachment_definition_enabled_instance_up{networks="any"}` and `network_attachment_definition_enabled_instance_up{networks="sriov"}` only
network_attachment_definition_instance_total will have all Unique network types configured for a pod via `network_attachment_definition`  example `network_attachment_definition_instance_total{networks="sriov"} =10 ` or ` ` `network_attachment_definition_instance_total{networks="mcvlan"}=20 `  or `network_attachment_definition_instance_total{networks="mcvlan, bridge"}=200 `
network_attachment_definition_enabled_instacne_up
network_attachment_definition_instance_total

brancz:
How come there are 3 of each? Is it one per node or are we running 3 replicas of this statically?

aneeshkp:
@brancz They run one per node, but collect the same metrics.

brancz:
Per node is too high of a cardinality than we can allow at this point unfortunately. We will need to roll up aggregates of this.

aneeshkp:
@brancz , will this  help if I cahnge the query to 
``` avg(network_attachment_definition_enabled_instance_up) by ( __name__, networks)
avg(network_attachment_definition_instance_total) by (__name__, networks) ```

**Comment by brancz** [comment]
Is `network_attachment_definition_instance_total` really a counter or a gauge (the `_total` suffix is reserved only for counters). An aggregation would be better yes, those are defined via recording rules by your component, similar to the ServiceMonitor object you create a PrometheusRule object.

**Comment by aneeshkp** [comment]
@brancz Thanks, I got the workaround with recording rules.  These are the gauge type , because the values go up and down. Now  I have the rules as shown below.
 ``` 
 - name: multus-admission-controller-monitor-service.rules
    rules:
    - expr: |
        avg without (instance, pod) (
          avg_over_time(network_attachment_definition_enabled_instance_up[1m])
        )
      record: cluster:network_attachment_definition_enabled_instance_up:rate1m  
    - expr: |
        avg without (instance, pod) (
          avg_over_time(network_attachment_definition_instance_total[1m])
        )
      record: cluster:network_attachment_definition_instance:rate1m 
``` 

and the output is 

instance_up

instance_rule

Let me if this is good for telemetry.

brancz:
Having an aggregation looks like we're going in the right direction, but it still seems like you want a sum not an average. I don't think you need an aggregation `_over_time`, a simple sum is enough. Also the `_total` suffix should be removed from that metric, as it's reserved for counters.

aneeshkp:
@brancz , Sum will give me wrong count of instances with annotations in the cluster. 
In this example, there are three metrics_pods running on 3 nodes and they are giving me the same set of metrics.
node 1 == says there are 3 pods with networks="any" in the cluster and node 2= also says there are 3 pods with networks="any"  in the cluster and node 3=also says there are 3 pods with networks="any"  in the cluster. Now if I do sum I will get 9, which is not correct. I think  I need an average in this case? 
Also, as you suggested  I will suffice with _sum instead of _total, I guess that's appropriate in this case?

brancz:
Ah got it, sorry my mistake. Average is ok then, maybe min/max would be better but that's your decision. `_sum` is a rather unusual suffix, typically metrics just have a plural name if they are a gauge and represent a "current count", like `container_memory_usage_bytes`. Whatever you find suitable with those constraints sounds good to me :slightly_smiling_face: .

aneeshkp:
Closing this PR and opening new with changes

PR #260, opened 2019-11-01
Add jaeger-oauth annotation on serviceaccount

metalmatze:
We need to add this annotation for the serving certs to appear.

/cc @aditya-konarde @brancz @kakkoyun @squat

### DIFF
--- a/jsonnet/jsonnetfile.json
+++ b/jsonnet/jsonnetfile.json
@@ -1,13 +1,13 @@
 {
-    "dependencies": [
-        {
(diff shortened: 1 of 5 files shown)


brancz:
It's getting a little ridiculous, we need to find a different solution for this I think. Kubernetes as the authn/authz provider just doesn't make sense.

/lgtm

metalmatze:
This just came back to haunt us already.......
For Jaeger to be deployed in production, we now need to deploy Telemeter...

PR #264, opened 2019-11-15
The content of benchmarking.pdf does not seem PDF.

adelton:
Addressing GitHub failing to display it with
Sorry, this file is invalid so it cannot be displayed.


squat:
Sorry, the file is actually a PDF, not a markdown file. Renaming it to .md won't make it viewable on GH. It contains the generated result of in-depth benchmarking tests. If you are interested in reading it, please clone the repo and open the file with your favorite PDF viewer :)

PR #268, opened 2019-12-03
jsonnet: Whitelist cluster:usage:* for telemetry

smarterclayton:
Allow developers to quickly add new global scoped, low cardinality
usage metrics that can be added without changing telemetry, CMO, or
observatorium. An e2e test will be added to Origin that watches for
excessive cardinality. We are explicitly trading off flexibility to
quickly iterate on new usage tracking metrics (and forcing consistent
structure) with direct control in order to reduce the cost to add
new usage tracking and backport it.


s-urbaniak:
/approve
as discussed, we should follow up asap with an e2e test, let's give this another set of eyes

smarterclayton:
Typo, applying label :)

LiliC:
LOL I will blame jetlag for that one as well :)

PR #271, opened 2019-12-05
cmd,pkg: eliminate gossip and prometheus fed

squat:
This commit eliminates the Prometheus federation portion of Telemeter
server. This effectively means that Telemeter server is now only an
authentication and validation proxy that forwards metrics via remote
write.

It also removes the gossip cluster feature of the project as this is no
longer. Previously it served to ensure that metrics only ended up on one
cluster node so that there would be no duplicate samples when federated
by Prometheus.

We can merge this once DataHub v1 has been taken down.
cc @metalmatze @kakkoyun


squat:
rebased on master to fix tests since https://github.com/openshift/telemeter/pull/313 merged

squat:
rebase on master again

PR #273, opened 2019-12-10
 cmd/telemeter-server,pkg/cache: cache auth resps

squat:
This commit adds new functionality to Telemeter server to enable it to
cache the responses to authentication requests made while handling
remote-write requests. The only currently implemented backend is
Memcached, because it is so simple, but any shared k-v would serve our
purposes.

cc @bwplotka @aditya-konarde @brancz @kakkoyun

### DIFF
--- a/cmd/telemeter-server/main.go
+++ b/cmd/telemeter-server/main.go
@@ -25,6 +25,7 @@ import (
 
 	oidc "github.com/coreos/go-oidc"
 	"github.com/oklog/run"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/spf13/cobra"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/clientcredentials"
@@ -36,6 +37,8 @@ import (
 	"github.com/openshift/telemeter/pkg/authorize/jwt"
 	"github.com/openshift/telemeter/pkg/authorize/stub"
 	"github.com/openshift/telemeter/pkg/authorize/tollbooth"
+	"github.com/openshift/telemeter/pkg/cache"
+	"github.com/openshift/telemeter/pkg/cache/memcached"
 	"github.com/openshift/telemeter/pkg/cluster"
 	telemeter_http "github.com/openshift/telemeter/pkg/http"
 	httpserver "github.com/openshift/telemeter/pkg/http/server"
@@ -86,6 +89,7 @@ func main() {
 		PartitionKey:       "_id",
 		Ratelimit:          4*time.Minute + 30*time.Second,
 		TTL:                10 * time.Minute,
+		MemcachedExpire:    24 * 60 * 60,
 	}
 	cmd := &cobra.Command{
 		Short:         "Aggregate federated metrics pushes",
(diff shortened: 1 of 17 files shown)


brancz:
/lgtm

Impressed how fast you threw this out there. Nice!

cmd/telemeter-server/main.go
```diff
@@ -86,6 +89,7 @@ func main() {
 		PartitionKey:       "_id",
 		Ratelimit:          4*time.Minute + 30*time.Second,
 		TTL:                10 * time.Minute,
+		MemcachedExpire:    24 * 60 * 60,
```
metalmatze: 24 hours seem a bit long. Are the JWTs valid for the same duration? I would have said 1 hour would be long already.

squat: That’s the current expiration time of the JWTs :/ so I think we should default to this for now just to keep traffic to tollbooth somewhat steady. We can of course override this with the flag at any point.

squat:
and totally agree about the e2e tests. Started getting into the weeds of it last night. We need:
* obsv/up generating metrics
* telemeter ingesting
* telemeter auth as a stub authorizer
* memcached for caching
* thanos receive ingesting
* thanos query for testing that metric made it through

pkg/authorize/tollbooth/tollbooth.go
```diff
@@ -75,3 +77,12 @@ func (a *authorizer) AuthorizeCluster(token, cluster string) (string, error) {
 
 	return response.AccountID, nil
 }
+
+// ExtractToken extracts the token from an auth request.
+// In the case of a request to Tollbooth, the token
+// is the entire contents of the request body.
+func ExtractToken(r *http.Request) (string, error) {
+	body, err := ioutil.ReadAll(r.Body)
```
bwplotka: Leaking body, let's close it.

pkg/cache/cache.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, errors.Wrap(err, "failed to generate key from request")
+	}
+
+	raw, ok, err := r.c.Get(key)
+	if err != nil {
+		r.cacheReadsTotal.WithLabelValues("error").Inc()
+		return nil, errors.Wrap(err, "failed to retrieve value from cache")
+	}
+
+	if !ok {
```
bwplotka: I would vote for hit being a main quick shortcut and miss a passing to next. The rule of thumb is that easier/shorter should go first.

pkg/cache/cache.go
```diff
@@ -0,0 +1,104 @@
...
+		r.cacheReadsTotal.WithLabelValues("miss").Inc()
+		resp, err := r.next.RoundTrip(req)
+		if err == nil && resp.StatusCode/200 == 1 {
+			// Try to cache the response but don't block.
+			defer func() {
+				raw, err := httputil.DumpResponse(resp, true)
+				if err != nil {
+					level.Error(r.l).Log("msg", "failed to dump response", "err", err)
```
bwplotka: It's kind of cache write error, right?

pkg/cache/memcached/memcached.go
```diff
@@ -0,0 +1,67 @@
...
+	i := memcache.Item{
+		Key:        key,
+		Value:      value,
+		Expiration: c.expiration,
+	}
+	return c.Client.Set(&i)
+}
+
+// hashKey hashes the given key to ensure that it is less than 250 bytes,
+// as Memcached cannot handler longer keys.
```
bwplotka:
```suggestion
// as Memcached cannot handle longer keys.
```

pkg/cache/cache.go
squat: yeah kind of, but not request was made to memcached. I'd like to ensure we have metrics to track failures to contact the cache.

bwplotka: Makes sense :+1:

pkg/authorize/handler.go
```diff
@@ -129,7 +132,8 @@ func NewHandler(logger log.Logger, client *http.Client, endpoint *url.URL, tenan
 
 		token, err := base64.StdEncoding.DecodeString(authParts[1])
 		if err != nil {
-			http.Error(w, "bad authorization header", http.StatusBadRequest)
+			level.Warn(logger).Log("msg", "failed to extract token", "err", err)
+			w.WriteHeader(http.StatusBadRequest)
```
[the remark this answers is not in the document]
squat: :/ nope, I think it's a mistake. I'll fix it here

PR #277, opened 2019-12-12
Enforce forwarding from v1 /upload to v2 receive hashring

metalmatze:
Until now we had transient errors for failures in forwarding the time series from our v1 /upload endpoint to the internal v2 Thanos receive hashring.

As we're defaulting to v2 soon and then want consistency, we should start enforcing those writes to fail. We didn't have any alerts firing on too many forwarding errors anyway.

/cc @brancz @bwplotka @kakkoyun @squat


brancz:
/lgtm
/hold

Feel free to remove hold when ready to deploy.

PR #278, opened 2019-12-12
pkg/cache: update memcached dns on interval

squat:
We need to be able to update the DNS for the Memcached DNS servers
to keep up with changes in the IPs of the servers. This implements a
simple time-based DNS update.

cc @metalmatze @brancz @kakkoyun

### DIFF
--- a/pkg/cache/memcached/memcached.go
+++ b/pkg/cache/memcached/memcached.go
@@ -1,8 +1,11 @@
 package memcached
 
 import (
+	"context"
 	"crypto/sha256"
 	"fmt"
+	"sync"
+	"time"
 
 	"github.com/bradfitz/gomemcache/memcache"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


pkg/cache/memcached/memcached.go
```diff
@@ -12,17 +15,38 @@ import (
 
 // cache is a Cacher implemented on top of Memcached.
 type cache struct {
-	*memcache.Client
+	client     *memcache.Client
 	expiration int32
+	sync.RWMutex
```
brancz: I'm not a huge fan of exposing these, I've seen this cause problems in the past as the methods happen to be callable from the outside.

brancz:
Just a small comment, otherwise lgtm

pkg/cache/memcached/memcached.go
squat: ack :+1:

PR #279, opened 2019-12-12
Receive requests are required to have partitionKey label

metalmatze:
Let's decode the receive requests and check if the partitionKey (for us mostly `_id`) is available in the request and otherwise reject the request.

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -67,6 +67,8 @@ func (h *Handler) Receive(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	defer r.Body.Close()
+
 	// Limit the request body size to a sane default
 	r.Body = http.MaxBytesReader(w, r.Body, requestLimit)
 
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -1,6 +1,7 @@
 package receive
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io/ioutil"
@@ -72,18 +73,6 @@ func (h *Handler) Receive(w http.ResponseWriter, r *http.Request) {
 	// Limit the request body size to a sane default
 	r.Body = http.MaxBytesReader(w, r.Body, requestLimit)
 
-	err := validateLabels(r, h.PartitionKey)
-	if err != nil {
-		level.Error(h.logger).Log("msg", "failed to validate labels in request", "err", err)
-
-		if err == ErrRequiredLabelMissing {
-			http.Error(w, err.Error(), http.StatusBadRequest)
-		} else {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-		}
-		return
-	}
-
 	ctx, cancel := context.WithTimeout(r.Context(), forwardTimeout)
 	defer cancel()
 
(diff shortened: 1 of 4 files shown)


pkg/receive/handler.go
```diff
@@ -71,3 +92,33 @@ func (h *Handler) Receive(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 }
+
+var ErrRequiredLabelMissing = fmt.Errorf("a required label is missing from the metric")
+
+// TODO: Make this a middleware eventually
```
squat: can we actually make this part of this same PR? Now that we are rewriting key pieces of telemetry, it's a good time to start writing more maintainable and composable funcs. Let's make a write-request validation middleware that can take a variable amount of `func(* prompb.WriteRequest) error` funcs and returns a handler. This way we can wrap the forwarder easily and separate concerns

pkg/receive/handler.go
```diff
@@ -43,7 +51,20 @@ func (h *Handler) Receive(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	defer r.Body.Close()
```
kakkoyun: @metalmatze Shouldn't we still make sure it's closed?

pkg/receive/handler.go
squat: can we still address this ^ before merging?

pkg/receive/handler.go
```diff
@@ -91,3 +103,49 @@ func (h *Handler) Receive(w http.ResponseWriter, r *http.Request) {
...
+		var wreq prompb.WriteRequest
+		if err := proto.Unmarshal(content, &wreq); err != nil {
+			http.Error(w, "failed to decode protobuf from body", http.StatusBadRequest)
+			return
+		}
+
+		found := false
+		for _, ts := range wreq.GetTimeseries() {
+			for _, l := range ts.GetLabels() {
+				if l.Name == h.PartitionKey {
```
squat: I think this logic is wrong. This currently checks if at least one time series on the request contains the label. We want to ensure that every time series has that label.

pkg/receive/handler.go
```diff
@@ -91,3 +103,49 @@ func (h *Handler) Receive(w http.ResponseWriter, r *http.Request) {
 	h.forwardRequestsTotal.WithLabelValues("success").Inc()
 	w.WriteHeader(resp.StatusCode)
 }
+
+var ErrRequiredLabelMissing = fmt.Errorf("a required label is missing from the metric")
+
+// ValidateLabels makes sure that the request's content contains the required partitionKey
+func (h *Handler) ValidateLabels(next http.Handler) http.HandlerFunc {
```
squat: It doesn't seem like this really needs to be a method of the handler struct. IMO it should be an function that accepts a label name and an next and returns an HandlerFunc. This would also eliminate the need for the constructor func NewHandler to accept an extra argument.

pkg/receive/handler.go
```diff
@@ -91,3 +103,49 @@ func (h *Handler) Receive(w http.ResponseWriter, r *http.Request) {
 	h.forwardRequestsTotal.WithLabelValues("success").Inc()
 	w.WriteHeader(resp.StatusCode)
 }
+
+var ErrRequiredLabelMissing = fmt.Errorf("a required label is missing from the metric")
```
squat: Let's add a comment to this exported type

pkg/receive/handler.go
```diff
@@ -61,6 +70,9 @@ func (h *Handler) Receive(w http.ResponseWriter, r *http.Request) {
 
 	defer r.Body.Close()
 
+	// Limit the request body size to a sane default
+	r.Body = http.MaxBytesReader(w, r.Body, requestLimit)
```
squat: This request limit should really be configurable via a command line argument so that we don't need to PR a golang change to bump the limit.

squat: We should also be able to disable the limi, eg if the limit is 0

squat: This limiter should also be its own Middleware i think. It doesn't really need to be couped to the logic of forwarding requests.

squat: This limiter should also be its own Middleware i think. It doesn't really need to be couped to the logic of forwarding requests.

pkg/receive/handler.go
metalmatze: I guess the reason I did it that way was to cut down on the amount of args being passed around on the caller side.
Happy to change it though :blush:

squat: yes, I think it's better; these two things should be entirely separate concerns that are not coupled. It will also allow us to generalize the label requirement more: rather than enforcing `_id=$token` we can enforce any arbitrary `$key=$label`

pkg/receive/handler.go
```diff
@@ -49,6 +56,8 @@ func NewHandler(logger log.Logger, forwardURL string, reg prometheus.Registerer)
 		reg.MustRegister(h.forwardRequestsTotal)
 	}
 
+	h.logger.Log("msg", "running receive handler")
```
squat: this log is a bit funny to me sine we're not running a goroutine here but creating a struct

pkg/receive/handler.go
```diff
@@ -91,3 +99,89 @@ func (h *Handler) Receive(w http.ResponseWriter, r *http.Request) {
 	h.forwardRequestsTotal.WithLabelValues("success").Inc()
 	w.WriteHeader(resp.StatusCode)
 }
+
+func LimitBodySize(limit int64, next http.Handler) http.HandlerFunc {
```
squat: could you add a short comment here

pkg/receive/handler.go
metalmatze: this might have been a test of mine with the logger not logging anything at first... :grin:

pkg/receive/handler.go
```diff
@@ -1,18 +1,25 @@
...
+	"github.com/gogo/protobuf/proto"
+	"github.com/golang/snappy"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/prometheus/prompb"
 
 	"github.com/openshift/telemeter/pkg/authorize"
 )
 
 const forwardTimeout = 5 * time.Second
+const RequestLimit = 15 * 1024 // based on historic Prometheus data with 6KB at most
```
squat: can we make this `DefaultRequestLimit`? Eventually we'll make the request limit overridable via flags, no?

metalmatze: Yes. :blush:

squat:
/lgtm

super cool :sunglasses:

PR #284, opened 2019-12-13
jsonnet/telemeter: fix setMemcached func

squat:
Currently if a container has no command, then the container will be
removed. This fixes the function.

cc @bwplotka @kakkoyun 


bwplotka:
Why is that needed? If container has no `command` default one would be taken from `Dockerfile` right?

squat:
but if there's no 'command' then the entire container is removed here.

PR #287, opened 2020-01-03
Dockerfile: update to Go 1.13

simonpasquier:


squat:
This will require bumping our ksonnet version to v4 , fixing resulting inconsistencies, and some testing. I can take care of that

metalmatze:
I am not sure if we can switch to StatefulSet `apps/v1` yet as I think it wasn't available on Kubernetes 1.11 yet... :roll_eyes:

LiliC:
> I am not sure if we can switch to StatefulSet apps/v1 yet as I think it wasn't available on Kubernetes 1.11 yet... 🙄

This only changes telemeter for 4.4 openshift version, not sure how does that effect k8s 1.11 that was shipped in openshift 3.11?

squat:
this was only causing errors on the benchmark test. so we bumped the API version on those manifests, not the actual server manifests used in prod. https://github.com/openshift/telemeter/pull/289

we should be able to rebase this PR and re-run now

PR #288, opened 2020-01-07
Consume authorize_url from Secret

maorfr:
This PR intends to take the authorization URL out of the template and into the telemeter-server Secret.

required for https://github.com/observatorium/configuration/pull/149

ultimately the purpose of this PR is to allow deploying ConfigMaps via upstream!

### DIFF
--- a/jsonnet/telemeter/server/kubernetes.libsonnet
+++ b/jsonnet/telemeter/server/kubernetes.libsonnet
@@ -135,7 +135,7 @@ local clusterPort = 8082;
       local secret = k.core.v1.secret;
 
       secret.new(secretName, {
-        authorize_url: std.base64($._config.telemeterServer.authorizeUrl),
+        authorize_url: '',
         oidc_issuer: std.base64($._config.telemeterServer.oidcIssuer),
         client_secret: std.base64($._config.telemeterServer.clientSecret),
         client_id: std.base64($._config.telemeterServer.clientID),
--- a/jsonnet/telemeter/server/kubernetes.libsonnet
+++ b/jsonnet/telemeter/server/kubernetes.libsonnet
@@ -13,7 +13,6 @@ local clusterPort = 8082;
     namespace: 'telemeter',
 
     telemeterServer+:: {
-      authorizeURL: 'https://api.openshift.com/api/accounts_mgmt/v1/cluster_registrations',
       replicas: 10,
       oidcIssuer: '',
       clientSecret: '',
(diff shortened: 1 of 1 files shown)


squat:
Hmm I’m pretty sure I actually remember having had this before and then switching to our current configuration. In any case, the authorization URL is not a secret and putting in there with the actual secret values makes it harder to maintain and debug IMO. Also, once nice thing about having this value outside of a secret is that any time that the argument is changed, the statefulset will be automatically rolled, whereas if the value in the secret changes we have to manually kick the pods.

maorfr:
thanks for your comments @squat!

this is not a secret, but it is non-public information IMO, as it discloses our architecture, even if only a small bit.

regarding kicking the pods - we can create an automation to recycle pods after a secret changed if we decide.

squat:
Adding more automation creates extra complication rather than use an already simple solution. The information is not “non public”: as long as there is an authorization URL flag at all then it is publicly known that we use an authorization server. The URL says nothing about the architecture other than “this thing that you already know exists is at this URL that you already can not access”. Also, the tollbooth auth flow/protocol is completely reversible from the code contained in this repo.

Please help me understand why this is a desired change.

maorfr:
the url says what is the authorization server. its one thing to know that it exists, and another to know where it is. we just want to add a bit of mystery to the game.

another reason for this change is that we want to allow you to deploy additional resources from upstream, such as ConfigMaps. To allow that, we looked at "what is non-public information" and how we can detect it, to prevent such info from residing in public repos. surprisingly, we found that in the majority of cases, a string starting with `https://` is sensitive information. As is this case, in my opinion.

This is the method that we are using at the moment to detect non-public information in public repos, and we plan to enable this check for this repo as well in the near future (https://gitlab.cee.redhat.com/service/app-interface/merge_requests/2554).

This change does not really make a huge difference IMO, so if we can get you on board that would be fantastic. for the record, we will not allow deploying ConfigMaps via this repo unless that app-interface PR is merged, so we have a shared interest here.

what do you think?

squat:
> for the record, we will not allow deploying ConfigMaps via this repo unless that app-interface PR is merged, so we have a shared interest here

:/ it doesn’t sound like this is really a choice


> we just want to add a bit of mystery to the game.

Any publicly accessible endpoint should *always* be treated as if everyone already knows it and may try to abuse it. Obscuring the URL is not a security measure IMO and should not be considered to make any progress towards protecting the resource.

jsonnet/telemeter/server/kubernetes.libsonnet
```diff
@@ -135,6 +135,7 @@ local clusterPort = 8082;
       local secret = k.core.v1.secret;
 
       secret.new(secretName, {
+        authorize_url: std.base64($._config.telemeterServer.authorizeUrl),
```
squat: With these changes Jsonnet will fail build this because there is no field named `authorize_url` in the config. How about leaving it as `''`?

maorfr: fixed

PR #293, opened 2020-01-10
tools.go: fix integration test by pinning deps

squat:
This PR switches to using the tools pattern for installing the go
build dependencies. This allows us to more easily pin our dependency on
UP so that we can fix the integration test.

This PR also bumps Telemeter to go 1.13

xref: https://github.com/openshift/telemeter/pull/291#issuecomment-572998496

cc @kakkoyun


simonpasquier:
`ci/prow/vendor` fails because the CI job is still using Go 1.12 which generates a different `vendor/modules.txt`.

squat:
Yes but it seems we have a chicken and egg problem because the PR in openshift/release to bump to use the go 1.13 image fails too

PR #294, opened 2020-01-10
test: Use specific version of UP to fix broken integration tests

kakkoyun:
This PR fixes integration tests that are broken after a dependency update.
The pipeline now uses a fixed version of the tool.


squat:
/lgtm
This is a temporary fix until we can get go 1.13 working in master, at which point we’ll switch to tools.go

kakkoyun:
xref: https://github.com/openshift/telemeter/pull/291 and https://github.com/openshift/telemeter/pull/293

PR #296, opened 2020-01-16
jsonnet/telemeter: Add recording rule for cluster availability

smarterclayton:
We have defined the OpenShift SLI based on whether the cluster
version operator is reporting a failure or not over a period of
time. To better support queries that attempt to aggregate availability
over classes of cluster, create a recording rule that is 1 when
a cluster is available and 0 when the cluster is reporting unhealthy.
If the cluster is not reporting any metrics it would have no data
at that point in time. Include the current version of the cluster
since it is implicit at the time of the recording rule.

This can be used in more complex aggregates by joining on cluster
id.


LiliC:
CI is having problems so lets wait until thats back. 🤷‍♀

PR #299, opened 2020-01-21
pkg/receive: return correct status code

squat:
status code, i.e. it does not overwrite the status code returned by the
downstream handler.

cc @metalmatze @brancz @kakkoyun @bwplotka


metalmatze:
Did you check what happens now if we actually have a gateway timeout?

squat:
If there's a gateway timeout, then `err` is not `nil`, so we hit this line of code:
https://github.com/openshift/telemeter/blob/95621237a985b078efbb26d5c4f12500a40e4cd3/pkg/receive/handler.go#L63

bwplotka:
As a side note this is something we can improve: Ideally we should use something like httputil.ReverseProxy which would handle all those things automatically.

PR #300, opened 2020-01-22
Fix remote write metrics

squat:
Today, we only track the number of errors encountered when forwarding requests from the v1 upload endpoint to thanos. We don't track the total amount of requests made from the v1 upload endpoint handler. We also don't track any requests made when forwarding requests from the v2 endpoint to thanos.

This PR adds these metrics to give us better visibility during another incident.

cc @metalmatze @brancz @bwplotka @kakkoyun

### DIFF
--- a/pkg/store/forward/forward.go
+++ b/pkg/store/forward/forward.go
@@ -26,21 +26,21 @@ const (
 
 var (
 	forwardSamples = prometheus.NewCounter(prometheus.CounterOpts{
-		Name: "telemeter_forward_samples_total",
-		Help: "Total amount of samples successfully forwarded",
+		Name: "telemeter_v1_forward_samples_total",
+		Help: "Total amount of successfully forwarded samples from v1 requests.",
 	})
 	forwardRequests = prometheus.NewCounterVec(prometheus.CounterOpts{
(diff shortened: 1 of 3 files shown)


pkg/store/forward/forward.go
```diff
@@ -26,27 +26,27 @@ const (
 
 var (
 	forwardSamples = prometheus.NewCounter(prometheus.CounterOpts{
-		Name: "telemeter_forward_samples_total",
-		Help: "Total amount of samples successfully forwarded",
-	})
-	forwardErrors = prometheus.NewCounter(prometheus.CounterOpts{
-		Name: "telemeter_forward_request_errors_total",
```
brancz: Do we have alerts for these metrics? If so we should probably change these as well?

squat: I'll investigate. In any case, there are no other references to those metrics in this repo, so we'lll need to make adjustments in either obs/cfg or the deployment repos

squat: ack, confirmed we don't have any alerts or recording rules referencing these metrics today. eventually we'll want to add some alerts based on the error rates tho

brancz: ack

PR #302, opened 2020-01-23
Makefile: execute built `jb` from correct path

cben:


cben:
This remove need for one tool.
One still has to obtain `jsonnet` binary.  I see `JSONNET_BIN` defaults to `$(BIN_DIR)/jsonnet` but nothing in makefile that builds it there...

kakkoyun:
@cben This should fix the issue that you've mentioned https://github.com/openshift/telemeter/pull/303

PR #304, opened 2020-01-24
Enable apiserver_current_inflight_requests in telemeter

tkashem:
Enable `apiserver_current_inflight_requests` in telemeter. 

`apiserver_current_inflight_requests` refers to maximal number of currently used in-flight request limit of this apiserver per request kind in last second. We want to have an idea of how loaded our api server(s) are globally.


tkashem:
@openshift/openshift-team-monitoring

brancz:
I assume the cardinality of this is bound to the number of apiservers (as in each apiserver exposes this)? In that case it would you get the same value out of this if we summed this up for the entire cluster and only sent that time series? That would turn this into a constant 1 time series per cluster rather than a variable number. If that’s not possible we need to look closer at the budgeting for this.

tkashem:
Hi @brancz 

> That would turn this into a constant 1 time series per cluster rather than a variable number. If that’s not possible we need to look closer at the budgeting for this.

Below is a sample of the gauge:
```apiserver_current_inflight_requests{endpoint="https",instance="10.128.0.42:6443",job="oauth-openshift",namespace="openshift-authentication",pod="oauth-openshift-6cb558fc44-bj52q",requestKind="mutating",service="oauth-openshift"}```

Yes, each kube-apiserver exposes this gauge. But I don't see any label that identifies the kube-apiserver instance. So I am assuming it's already aggregated at the source. 

The cardinality of this gauge is bound to the label `requestKind` which has two fixed values, either `readOnly` or `mutating`. We would like to keep this distinction, if possible. 

The rest of the labels I assume are added by default and will be dropped to turn it into a two time series, one for `readOnly` and the other `mutating` per cluster.

**Comment by tkashem** [comment]
@brancz I am sending 2 time series per cluster now. `sum(apiserver_current_inflight_requests) by(requestKind)`. Let me know if this is okay.

File: docs/data-collection.md [hunk 1]
```diff
@@ -162,6 +162,10 @@ For the OpenShift 4 Developer Preview we will be sending back these exact attrib
   // cam_app_workload_migrations tracks number of app workload migrations
   // by current state. Tracked migration states are idle, running, completed, and failed.
   '{__name__="cam_app_workload_migrations"}',
+  // apiserver_current_inflight_requests refers to maximal number of currently used
+  // inflight request limit of an apiserver per request kind in last second.
+  // We want to have an idea of how loaded our api server(s) are globally.
+  '{__name__="sum(apiserver_current_inflight_requests) by(requestKind)"}',
```
LiliC: Can we rename the label `requestKind` to just `kind`?

docs/data-collection.md
```diff
@@ -162,6 +162,10 @@ For the OpenShift 4 Developer Preview we will be sending back these exact attrib
   // cam_app_workload_migrations tracks number of app workload migrations
   // by current state. Tracked migration states are idle, running, completed, and failed.
   '{__name__="cam_app_workload_migrations"}',
+  // apiserver_current_inflight_requests refers to maximal number of currently used
+  // inflight request limit of an apiserver per request kind in last second.
```
LiliC: In last second? What do you mean here?

docs/data-collection.md
tkashem: This metric is exposed by upstream `kube-apiserver`.

brancz: this doesn't work, you need to have a separate recording rule that aggregates this in your namespace and then just whitelist the recording rule here

docs/data-collection.md
tkashem: updated the comment, let me know if this is clear.

docs/data-collection.md
```diff
@@ -162,6 +162,11 @@ For the OpenShift 4 Developer Preview we will be sending back these exact attrib
   // cam_app_workload_migrations tracks number of app workload migrations
   // by current state. Tracked migration states are idle, running, completed, and failed.
   '{__name__="cam_app_workload_migrations"}',
+  // cluster:apiserver_current_inflight_requests:sum:max_over_time:1m gives maximum number of requests in flight
+  // over a 1-minute window. This metric is a constant 4 time series that monitors concurrency of kube-apiserver and
+  // openshift-apiserver with request type which can be either 'mutating' or 'readonly'.
+  // We want to have an idea of how loaded our api server(s) are globally.
+  '{__name__="cluster:apiserver_current_inflight_requests:sum:max_over_time:1m"}',
```
brancz: What’s the scrape interval for the apiserver? 1m range query might fail if there is not at least 3x scrape intervals in 1min. I would suggest to bump it at least to 2m if not 5m.

docs/data-collection.md
```diff
@@ -162,6 +162,11 @@ For the OpenShift 4 Developer Preview we will be sending back these exact attrib
   // cam_app_workload_migrations tracks number of app workload migrations
   // by current state. Tracked migration states are idle, running, completed, and failed.
   '{__name__="cam_app_workload_migrations"}',
+  // cluster:apiserver_current_inflight_requests:sum:max_over_time gives maximum number of requests in flight
+  // over a 1-minute window. This metric is a constant 4 time series that monitors concurrency of kube-apiserver and
+  // openshift-apiserver with request type which can be either 'mutating' or 'readonly'.
+  // We want to have an idea of how loaded our api server(s) are globally.
+  '{__name__="cluster:apiserver_current_inflight_requests:sum:max_over_time"}',
```
brancz: having the range window in the name was correct, but the docs still say 1m, was this actually changed?

tkashem:
@brancz I addressed your feedback on the related PRs https://github.com/openshift/cluster-kube-apiserver-operator/pull/737 and https://github.com/openshift/cluster-openshift-apiserver-operator/pull/307.

If you have any further suggestions on this PR please let me know.

tkashem:
new process to send metric to telemeter, closing this PR. New PR https://github.com/openshift/cluster-monitoring-operator/pull/643

PR #298, opened 2020-01-17
Capture total CAM app workload migrations by migration state.

djwhatle:
Add metric gathering for `cam_app_workload_migration_total`

- Metric was originally added to mig-controller in https://github.com/fusor/mig-controller/pull/379 , got feedback from Clayton
- Metric name was adjusted after consulting with @LiliC and @smarterclayton in https://github.com/fusor/mig-controller/pull/385
- Metric was added to mig-operator in https://github.com/fusor/mig-operator/pull/193

Metric is intended to help us understand at a high level how customers are using app workload migration tooling (CAM) to move apps and PV data between OpenShift clusters, especially when moving from OpenShift 3->4.

PR #307, opened 2020-02-13
Bug 1803106: Backport cam_app_workload_migrations metric to release-4.3

djwhatle:
Follow up to https://github.com/openshift/telemeter/pull/298, backport of `cam_app_workload_migrations` metric was requested by Clayton per https://github.com/openshift/telemeter/pull/298#issuecomment-577158997

### DIFF
--- a/docs/data-collection.md
+++ b/docs/data-collection.md
@@ -162,6 +162,9 @@ For the OpenShift 4 Developer Preview we will be sending back these exact attrib
   // cluster:network_attachment_definition_enabled_instance_up  informs (1 or 0) if the cluster has
   //at least max of one instance with  k8s.v1.cni.cncf.io/networks annotation, labelled by networks (any or sriov).
   '{__name__="cluster:network_attachment_definition_enabled_instance_up:max"}',
+  // cam_app_workload_migrations tracks number of app workload migrations
+  // by current state. Tracked migration states are idle, running, completed, and failed.
+  '{__name__="cam_app_workload_migrations"}',
 ]
 ```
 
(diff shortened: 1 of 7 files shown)


djwhatle:
@s-urbaniak what needs to happen to the BZ linked to this PR before we can merge? QE has verified this functionality and commented on the BZ.

lilic:
The linked bugzilla has to be in one of the states as the bot said:
> expected the bug to be in one of the following states: NEW, ASSIGNED, ON_DEV, POST, POST, but it is MODIFIED instead

So move the relevant 4.3 bugzilla to ASSIGNED and refresh the bug with the command. Hope that makes sense.

Note: There are some docs around this in openshift mojo page on how to backport process works, if you search on mojo I think there is.

djwhatle:
> So move the relevant 4.3 bugzilla to ASSIGNED and refresh the bug with the command. Hope that makes sense.

Done, I moved the BZ to ASSIGNED. QE has verified that these CAM metrics show up in the web UI. What should they move the BZ to? I didn't see that level of detail specified in the mojo doc that I found on the OpenShift backport process.

lilic:
> A bit confused by this whole process since CAM isn't formally tied to OpenShift releases, and runs across any OpenShift version from 3.7 to 4.4. Maybe we don't fit the process as well as some other operators.

You still require a bugzilla to merge things into openshift itself.

lilic:
Do you mind rebasing now, then we are good to go, thanks.

djwhatle:
@lilic rebased 👍

PR #311, opened 2020-02-24
jsonnet: Add recording rule to precalculate non red hat, non ibm clusters

brancz:
This recording rule will pre-calculate one of the more expensive parts of queries that we perform often in similar ways across a lot of queries.

@smarterclayton @metalmatze @s-urbaniak @lilic @bwplotka @squat

### DIFF
--- a/jsonnet/telemeter/rules.libsonnet
+++ b/jsonnet/telemeter/rules.libsonnet
@@ -35,6 +35,12 @@
                 topk by (_id) (1, max by (_id, managed, ebs_account, internal) (label_replace(label_replace((subscription_labels{support=~"Standard|Premium|Layered"} * 0 + 1) or subscription_labels * 0, "internal", "true", "email_domain", "redhat.com|(.*\\.|^)ibm.com"), "managed", "", "managed", "false")) + on(_id) group_left(version) (topk by (_id) (1, 0*cluster_version{type="current"})))
               |||,
             },
+            {
+              record: 'subscription_labels:not_redhat_not_ibm',
+              expr: |||
+                topk by (_id) (1, 0 *subscription_labels{email_domain!~"redhat.com|(^|.*\\.)ibm.com"})
+              |||,
+            },
           ],
         },
       ],
--- a/jsonnet/telemeter/rules.libsonnet
+++ b/jsonnet/telemeter/rules.libsonnet
@@ -38,7 +38,24 @@
             {
               record: 'subscription_labels:not_redhat_not_ibm',
               expr: |||
-                topk by (_id) (1, 0 *subscription_labels{email_domain!~"redhat.com|(^|.*\\.)ibm.com"})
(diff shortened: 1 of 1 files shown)


jsonnet/telemeter/rules.libsonnet
```diff
@@ -35,6 +35,12 @@
                 topk by (_id) (1, max by (_id, managed, ebs_account, internal) (label_replace(label_replace((subscription_labels{support=~"Standard|Premium|Layered"} * 0 + 1) or subscription_labels * 0, "internal", "true", "email_domain", "redhat.com|(.*\\.|^)ibm.com"), "managed", "", "managed", "false")) + on(_id) group_left(version) (topk by (_id) (1, 0*cluster_version{type="current"})))
               |||,
             },
+            {
+              record: 'subscription_labels:not_redhat_not_ibm',
+              expr: |||
+                topk by (_id) (1, 0 *subscription_labels{email_domain!~"redhat.com|(^|.*\\.)ibm.com"})
```
lilic: This is what I stole from Clayton IIRC, and maybe we could add another recording rule for the first part as that is something they use from what I could tell:
```
(
  ((time() - topk by (_id) (1,
    label_replace(
      label_replace(cluster_version{type="current"}, "version", "4.$1-0.$2", "version", "4\\.(\\d+\\.\\d+)-0.(ci|nightly|okd).*"),
      "version",
      "ci-pull-requests",
      "version",
      "0.0.1-.*"
    )
  )))
  + on (_id) group_left() (0 * topk by (_id) (1, subscription_labels{email_domain!~"redhat.com|(^|.*\\.)ibm.com"})) + 0
))
```

brancz:
> A bit weird is the 0* before subscription_labels but that has some dashboarding reasons, I guess.

Yes it's so we can do a join in a sum like `+ on(_id) group_left() (subscription_labels:not_redhat_not_ibm)`.

jsonnet/telemeter/rules.libsonnet
brancz: that's also a good one, let me add that now already

jsonnet/telemeter/rules.libsonnet
```diff
@@ -35,6 +35,29 @@
                 topk by (_id) (1, max by (_id, managed, ebs_account, internal) (label_replace(label_replace((subscription_labels{support=~"Standard|Premium|Layered"} * 0 + 1) or subscription_labels * 0, "internal", "true", "email_domain", "redhat.com|(.*\\.|^)ibm.com"), "managed", "", "managed", "false")) + on(_id) group_left(version) (topk by (_id) (1, 0*cluster_version{type="current"})))
               |||,
             },
+            {
+              record: 'subscription_labels:not_redhat_not_ibm',
```
smarterclayton: I don't think we need this because we have `id_version_ebs_account_internal:cluster_subscribed`

brancz: agreed, closing

PR #313, opened 2020-02-27
Move benchmark to remote write

squat:
This PR migrates the Telemeter server benchmark tests from the
Prometheus to the remote-write architecture. This will allow us to
completely eliminate the Prometheus storage stack, including manifests
and go code.


squat:
unblocks: https://github.com/openshift/telemeter/pull/271

squat:
This is failing because the benchmark test is not using the telemeter image built from the code in the PR. It is using `quay.io/openshift/origin-telemeter:v4.0`, which does not support the `--forward-url` flag.

We need to update the tests to use the telemeter image built from the source.

brancz:
As a follow up, let's remove the prometheus operator dependency in the jsonnetfile, as this is not required anymore, coincidentally that will make bumping kube-prometheus _a lot_ easier in cluster-monitoring-operator! :tada:

PR #314, opened 2020-03-03
vendor: bump thanos dep to v0.11.0 for tests

squat:
cc @metalmatze


squat:
ack, should not be a significant issue in these integration tests :+1:

PR #315, opened 2020-03-03
test/integration: fix already used port race

squat:
Occasionally, our integration tests are flaking because a process
listening on a certain port in the first integration test has not exited
before another process tries to listen on that port in the second
integration test. This commit gives the processes in the second
integration test a separate set of ports to ensure that the integration
tests do not conflict with one-another.

cc @metalmatze

### DIFF
--- a/test/integration-v2.sh
+++ b/test/integration-v2.sh
@@ -9,17 +9,17 @@ set -euo pipefail
 result=1
 trap 'kill $(jobs -p); exit $result' EXIT
 
-( ./authorization-server localhost:9001 ./test/tokens.json ) &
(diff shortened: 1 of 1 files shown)


test/integration-v2.sh
```diff
@@ -9,41 +9,41 @@ set -euo pipefail
 result=1
```
bwplotka: I would at some point rewrite it to https://github.com/thanos-io/thanos/blob/master/test/e2e/e2ethanos/services.go#L154

squat: yeah, I've been thinking about this for a while; this repo has a lot of space for refactoring, both the tests and the actual packages. These recent changes will at least unblock us to keep cleaning up

PR #319, opened 2020-03-17
Bug 1812719: jsonnet/telemeter: set cpu requests, unset limits

s-urbaniak:
This sets CPU requests to what I measured during an e2e run.
At the same time it unsets limits as this was recommended for OpenShift workloads.


squat:
@s-urbaniak you'll need to rebase onto #317 in a second when that merges

PR #321, opened 2020-03-24
Move all routes and handlers into scoped blocks

metalmatze:
This helps making individual dependencies more clear and should show
better how to refactor going forward.

/cc @squat

### DIFF
--- a/cmd/telemeter-server/main.go
+++ b/cmd/telemeter-server/main.go
@@ -20,7 +20,7 @@ import (
 	"strings"
 	"time"
 
-	oidc "github.com/coreos/go-oidc"
+	"github.com/coreos/go-oidc"
 	"github.com/oklog/run"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


cmd/telemeter-server/main.go
```diff
@@ -275,212 +275,218 @@ func (o *Options) Run() error {
...
-		transforms.With(metricfamily.NewLabel(o.Labels, nil))
-	}
-	transforms.With(metricfamily.NewElide(o.ElideLabels...))
+			auth := jwt.NewAuthorizeClusterHandler(o.Logger, o.PartitionKey, o.TokenExpireSeconds, signer, o.RequiredLabels, clusterAuth)
+			validator := validate.New(o.PartitionKey, o.LimitBytes, 24*time.Hour, time.Now)
 
-	server := httpserver.New(o.Logger, store, validator, transforms)
-	receiver := receive.NewHandler(o.Logger, o.ForwardURL, prometheus.DefaultRegisterer)
+			var store store.Store
+			u, err := url.Parse(o.ForwardURL)
```
squat: this is not really specific to the v1 routes since it's used by v2 storage too, right? maybe we can move this flag parsing to ~ line 277 with some of the other parsing

cmd/telemeter-server/main.go
```diff
@@ -275,212 +275,218 @@ func (o *Options) Run() error {
...
+			for i := 0; i < len(o.Whitelist); {
+				s := strings.TrimSpace(o.Whitelist[i])
+				if len(s) == 0 {
+					o.Whitelist = append(o.Whitelist[:i], o.Whitelist[i+1:]...)
+					continue
+				}
+				o.Whitelist[i] = s
+				i++
+			}
+			whitelister, err := metricfamily.NewWhitelist(o.Whitelist)
```
squat: TODO we need to eventually port this functionality into the remote write api as well

metalmatze: We should be able to make this a middleware just like everything else, right?

squat: yes exactly

PR #322, opened 2020-03-26
jsonnet: Ensure env var is actually added

brancz:
@metalmatze @squat 

cc @maorfr

### DIFF
--- a/jsonnet/telemeter/server/kubernetes.libsonnet
+++ b/jsonnet/telemeter/server/kubernetes.libsonnet
@@ -43,7 +43,6 @@ local internalPort = 8081;
       local podLabels = { 'k8s-app': 'telemeter-server' };
       local tlsMount = containerVolumeMount.new(tlsVolumeName, tlsMountPath);
       local tlsVolume = volume.fromSecret(tlsVolumeName, tlsSecret);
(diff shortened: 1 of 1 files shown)


jsonnet/telemeter/server/kubernetes.libsonnet
```diff
@@ -96,7 +96,7 @@ local internalPort = 8081;
         container.mixin.resources.withLimitsMixin($._config.telemeterServer.resourceLimits) +
         container.mixin.resources.withRequestsMixin($._config.telemeterServer.resourceRequests) +
         container.withVolumeMounts([tlsMount]) +
-        container.withEnv([oidcIssuer, clientSecret, clientID]) + {
+        container.withEnv([name, authorizeURL, oidcIssuer, clientSecret, clientID]) + {
```
squat: I think we can actually get rid of this environment variable entirely. It was previously used to give each pod a unique ID in the gossip hashring but I think it's not longer needed.

brancz: ack, I thought so

brancz: done

PR #323, opened 2020-04-02
Add alerts and cluster_operator_conditions recording rules for OCM to query

nimrodshn:
cc: @brancz @smarterclayton @igoihman @cben PTAL

### DIFF
--- a/jsonnet/telemeter/rules.libsonnet
+++ b/jsonnet/telemeter/rules.libsonnet
@@ -35,6 +35,18 @@
                 topk by (_id) (1, max by (_id, managed, ebs_account, internal) (label_replace(label_replace((subscription_labels{support=~"Standard|Premium|Layered"} * 0 + 1) or subscription_labels * 0, "internal", "true", "email_domain", "redhat.com|(.*\\.|^)ibm.com"), "managed", "", "managed", "false")) + on(_id) group_left(version) (topk by (_id) (1, 0*cluster_version{type="current"})))
               |||,
             },
+            {
+              record: 'id:critical_alerts_firing:count',
+              expr: |||
+                count by (_id) (alerts{alertstate='firing', severity='critical'})
+              |||,
+            },
+            {
+              record: 'id:cluster_operator_failing:count',
+              expr: |||
+                count by (_id) (cluster_operator_conditions{condition="Failing"})
+              |||,
+            },
           ],
         },
       ],


jsonnet/telemeter/rules.libsonnet
```diff
@@ -35,6 +35,18 @@
                 topk by (_id) (1, max by (_id, managed, ebs_account, internal) (label_replace(label_replace((subscription_labels{support=~"Standard|Premium|Layered"} * 0 + 1) or subscription_labels * 0, "internal", "true", "email_domain", "redhat.com|(.*\\.|^)ibm.com"), "managed", "", "managed", "false")) + on(_id) group_left(version) (topk by (_id) (1, 0*cluster_version{type="current"})))
               |||,
             },
+            {
+              record: 'id:alerts:count',
```
aditya-konarde, line 39: It may be misleading to call this recorded metric `id:alerts:count` as it records specifically only the alerts that are in `firing` state and with a severity of `critical`

jsonnet/telemeter/rules.libsonnet
```diff
@@ -35,6 +35,18 @@
...
+              expr: |||
+                count by (_id) (alerts{alertstate='firing', severity='critical'})
+              |||,
+            },
+            {
+              record: 'id:cluster_operator_conditions:count',
```
aditya-konarde, line 45: Same as above, possibly may want to change the recording rule name here to clarify that these are only `Failing` operators. See https://github.com/openshift/telemeter/blob/d7dcf4c7a2ba7949a06c36dc12e8e7591d29b77f/jsonnet/telemeter/rules.libsonnet#L15-L18 for an example where recorded metric name provides a hint on what the label selectors used were.

jsonnet/telemeter/rules.libsonnet
cben: Would {alertstate='firing', severity='critical'} labels be copied to the new metric?
If yes it _might_ be ok, similar to how only `alerts{alertstate='firing'}` subset is sent to telemetry but the metric is still  called just `alerts`.

(I don't have a qualified opinion, don't know much about metric naming conventions)

brancz:
Slightly updated comment from what I commented on gitlab:

Performance and resulting amount of time-series wise, these seem ok. It would be good to get Clayton's input as well though as it's a bit of a cross cutting concern, and this could just as easily be calculated in cluster, as there is no cross-cluster correlation happening here, which is a characteristic of most other use cases of the read path. Wdyt @smarterclayton ?

cben:
What if we go further and add a recording rule for overall "is cluster healthy for OCM purposes", combining all of nodes down/critical alerts/failing operators/high cpu/men into a single Boolean?

(Frederic's question of computing it in-cluster applies to this as well, but for a high-level concept like that I think the definition is gonna evolve, and doing it in-cluster would not give us ability to roll out a new definition for all clusters)

jsonnet/telemeter/rules.libsonnet
metalmatze: They won't. But you can add them to the labels of the recording rules just as well:

```
{
              record: 'id:alerts:count',
              expr: |||
                count by (_id) (alerts{alertstate='firing', severity='critical'})
              |||,
              labels: {
                alertstate: "firing",
                severity: "critical",
              }
            },

nimrodshn:
> What if we go further and add a recording rule for overall "is cluster healthy for OCM purposes", combining all of nodes down/critical alerts/failing operators/high cpu/men into a single Boolean?

@cben I am not opposed for this whatsoever but keep in mind we would need to alter the UI implementation to use the new `recording` rule as well as it currently doesn't account for node up / down but only for the latter conditions (i.e. `count(critical_alerts_firing) > 0 or count(operators_condition_failing) > 0 or cpu_usage exceed 95% or mem_usage exceed 95%`)

brancz:
Can we take a quick step back here and answer why we even need these rules? The evaluations for _all_ clusters is already fast, so why can't we query this on demand whenever OCM displays this?

PR #325, opened 2020-04-03
Refactor v1 HTTP to use middlewares and http.HandlerFuncs

metalmatze:
Now that Telemeter doesn't have the hashring anymore, we can go ahead and get rid of the `store.Store` interface, allowing us to move to simpler architecture. 
With this PR we're moving to a `pkg/server` that contains multiple HTTP middlewares and http.HandlerFuncs.

/cc @squat @kakkoyun @krasi-georgiev @bwplotka @brancz


cmd/telemeter-server/main.go
```diff
@@ -325,7 +322,12 @@ func (o *Options) Run() error {
 		})
 	}
 	{
-		external := http.NewServeMux()
+		external := chi.NewRouter()
```
squat, line 325: nit: above we call the router `r` and the mux `internal` and here we are following a different pattern. let's keep it consistent.

cmd/telemeter-server/main.go
```diff
@@ -403,17 +405,12 @@ func (o *Options) Run() error {
 			}
 
 			auth := jwt.NewAuthorizeClusterHandler(o.Logger, o.PartitionKey, o.TokenExpireSeconds, signer, o.RequiredLabels, clusterAuth)
-			validator := validate.New(o.PartitionKey, o.LimitBytes, 24*time.Hour, time.Now)
+			//validator := validate.New(o.PartitionKey, o.LimitBytes, 24*time.Hour, time.Now)
```
squat, line 408: are we intentionally ignoring this? do we just want to delete the line?

pkg/server/forward.go
```diff
@@ -51,68 +53,74 @@ func init() {
...
-func New(logger log.Logger, url *url.URL, next store.Store) *Store {
-	return &Store{
-		next:   next,
-		url:    url,
-		client: &http.Client{},
-		logger: log.With(logger, "component", "store/forward"),
-	}
-}
+	return func(w http.ResponseWriter, r *http.Request) {
+		partitionKey, ok := PartitionFromContext(r.Context())
```
squat, line 62: let's rename this `partitionKey` thing. It's a relic from the `partitioned` metrics type but it's a  weird name IMO. I think something like `ID` or anything would be much more clear

pkg/server/forward.go
```diff
@@ -51,68 +53,74 @@ func init() {
...
 		if err != nil {
-			return err
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
 		}
 
 		if len(timeseries) == 0 {
-			level.Info(s.logger).Log("msg", "no time series to forward to receive endpoint")
-			return nil
+			level.Info(logger).Log("msg", "no time series to forward to receive endpoint")
```
squat, line 93: Should we also respond with an HTTP error here?

pkg/server/ratelimited.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"fmt"
+	"net/http"
+	"sync"
+	"time"
+
+	"golang.org/x/time/rate"
+)
+
+// ErrWriteLimitReached is an error that is returned when a cluster has send too many requests.
```
squat, line 12: nit: s/send/sent/

pkg/server/snappy.go
```diff
@@ -0,0 +1,31 @@
...
+		reader := r.Body
+
+		if r.Header.Get("Content-Encoding") == "snappy" {
+			body, err := ioutil.ReadAll(r.Body)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+			}
+			defer r.Body.Close()
+
+			payload, _ := snappy.Decode(nil, body)
```
squat, line 23: I suggest we handle this error just in case

pkg/server/snappy.go
```diff
@@ -0,0 +1,31 @@
...
+			reader = ioutil.NopCloser(bytes.NewBuffer(payload))
+		}
+
+		r.Body = reader
```
squat, line 27: I think we can put this line and line 14 inside the `if` block, no? otherwise it's funny following why we are changing the body when the encoding is not snappy

cmd/telemeter-server/main.go
metalmatze: commented out during refactoring for reference. The `validate.New` func doesn't even exist anymore. Will remove.

pkg/server/forward.go
metalmatze: Yeah, I had the same thought during refactoring. I would have gone for `clusterID`. That's what it's ultimately is, right?

pkg/server/snappy.go
metalmatze: Seems like I've forgotten :upside_down_face:

test/e2e/forward_test.go
```diff
@@ -71,19 +68,20 @@ func TestForward(t *testing.T) {
 	}
 	var telemeterServer *httptest.Server
 	{
+		logger := log.NewNopLogger()
+
 		labels := map[string]string{"cluster": "test"}
-		validator := validate.New("cluster", 0, 0, time.Now)
+		//validator := validate.New("cluster", 0, 0, time.Now)
```
squat, line 74: can we simply delete this line?

pkg/server/forward.go
squat: yes exactly

cmd/telemeter-server/main.go
metalmatze: The problem is that the above still needs the internal to begin with. We need to refactor `telemeter_http.DebugRoutes` etc in another PR and then can have consistency.

pkg/server/forward.go
metalmatze: It was a transient error before. We simply ignored those requests and returned 200. I don't really care, honestly.

cmd/telemeter-server/main.go
squat: no I simply mean we should call the router `r` in both cases

pkg/server/forward.go
```diff
@@ -121,38 +129,24 @@ func (s *Store) WriteMetrics(ctx context.Context, p *store.PartitionedMetrics) e
...
 		meanDrift := timeseriesMeanDrift(timeseries, time.Now().Unix())
 		if math.Abs(meanDrift) > 10 {
-			level.Info(s.logger).Log("msg", "mean drift from now for clusters", "partitionkey", p.PartitionKey, "drift", fmt.Sprintf("%.3fs", meanDrift))
+			level.Info(logger).Log("msg", "mean drift from now for clusters", "clusterID", clusterID, "drift", fmt.Sprintf("%.3fs", meanDrift))
 		}
 
 		if resp.StatusCode/100 != 2 {
-			return fmt.Errorf("response status code is %s", resp.Status)
+			// surfacing upstreams error to our users too
+			http.Error(w, fmt.Errorf("response status code is %s", resp.Status).Error(), resp.StatusCode)
```
squat, line 137: This isn't correct IMO. If telemeter-server gets a 400 from the upstream then it is not the user's fault that there is an auth error, but rather an internal server error. The user should not receive a 400 for an improperly configured server. Instead the user should get a 500.

pkg/server/forward.go
```diff
@@ -51,68 +53,74 @@ func init() {
...
-type Store struct {
-	next   store.Store
-	url    *url.URL
-	client *http.Client
-	logger log.Logger
-}
+// ForwardHandler gets a request containing metric families and
+// converts it to a remote write request forwarding it to the upstream at fowardURL.
+func ForwardHandler(logger log.Logger, forwardURL *url.URL) http.HandlerFunc {
+	client := http.Client{Timeout: 10 * time.Second}
```
squat, line 59: This timeout is 10 seconds but the context for the forward request times out in 5 seconds. Why are they different? and why do we need to timeout mechanisms? I think this will lead to confusion about which one really applies. I would say let's keep just one of the timeout mechanisms.

pkg/server/instrument.go
```diff
@@ -36,7 +36,8 @@ func init() {
 	prometheus.MustRegister(requestDuration, requestSize, requestsTotal)
 }
 
-func NewInstrumentedHandler(handlerName string, next http.Handler) http.Handler {
+// InstrumentedHandler is a HTTP middleware that monitors HTTP requests and responses.
```
squat, line 39: nit: ... is an HTTP ...

pkg/server/ratelimited.go
```diff
@@ -0,0 +1,62 @@
...
+	"time"
+
+	"golang.org/x/time/rate"
+)
+
+// ErrWriteLimitReached is an error that is returned when a cluster has sent too many requests.
+type ErrWriteLimitReached string
+
+func (e ErrWriteLimitReached) Error() string {
+	return fmt.Sprintf("write limit reached for key %q", string(e))
```
squat, line 16: maybe we should rename `key` to `id` or `cluster` now

pkg/server/snappy_test.go
```diff
@@ -0,0 +1,71 @@
...
+		resp, err := s.Client().Do(req)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if resp.StatusCode != http.StatusOK {
+			t.Errorf("expected %d and got %d", http.StatusOK, resp.StatusCode)
+		}
+	}
+	//{
```
squat, line 56: do we want to re-enable this test or should we remove it simply?

pkg/server/validator.go
```diff
@@ -0,0 +1,143 @@
...
+	return context.WithValue(ctx, clusterIDCtx, clusterID)
+}
+
+// ClusterIDFromContext returns the clusterID from the context.
+func ClusterIDFromContext(ctx context.Context) (string, bool) {
+	p, ok := ctx.Value(clusterIDCtx).(string)
+	return p, ok
+}
+
+type clusterIDCtxType int
```
squat, line 51: let's put these definitions at the top of the file to follow convention

pkg/server/validator.go
```diff
@@ -0,0 +1,143 @@
...
+			return
+		}
+		defer r.Body.Close()
+
+		var transforms metricfamily.MultiTransformer
+		transforms.With(metricfamily.NewErrorOnUnsorted(true))
+		transforms.With(metricfamily.NewRequiredLabels(client.Labels))
+		transforms.With(metricfamily.TransformerFunc(metricfamily.DropEmptyFamilies))
+
+		if limitBytes > 0 {
```
squat, line 78: hmm this is funny, shouldn't this come before the `ioutil.ReadAll` on line 66?

pkg/server/validator.go
```diff
@@ -0,0 +1,143 @@
...
+			r.Body = reader.NewLimitReadCloser(r.Body, limitBytes)
```
squat, line 79: what effect does this have? it seems like the entire body was already read into a new buffer called `body` on line 66

pkg/server/validator_test.go
```diff
@@ -0,0 +1,180 @@
...
+			}))
+			next.ServeHTTP(w, r)
+		}
+	}
+
+	s := httptest.NewServer(
+		fakeAuthorizeHandler(
+			Validate(metricfamily.MultiTransformer{}, time.Hour, 512*1024, time.Now,
+				func(w http.ResponseWriter, r *http.Request) {
+					// TODO: Make the check proper to changing timestamps?
```
squat, line 43: Should we delete or re-enable this?

pkg/server/forward.go
metalmatze: Right. That makes sense. Keep in mind that this is the request send from Telemeter to Thanos Receivers. I guess it still holds true then too?

pkg/server/forward.go
metalmatze: You're right. I added this as another safeguard, but we don't really need it here. Removing.

pkg/server/ratelimited.go
metalmatze: `write limit reached for cluster with id`? :blush:

brancz:
This is entirely code that is not shipped in OpenShift, but needed for the telemetry server infrastructure. All CI green. Merging.

PR #329, opened 2020-04-28
Create a request logger per middleware / handler

metalmatze:
For every middleware or handler we create a new logger and get the request ID from the request.Context. However, adding a new key-value-pair to the logger per request, we need a new request logger instance. That why we now have `rlogger` to create the request logger.

/cc @brancz @kakkoyun @squat


brancz:
Once again merging, as this is 100% telemetry server side, and we're fixing a production system :)

PR #332, opened 2020-05-05
telemeter-server: Always set OpenShift telemetry tenant ID

brancz:
@metalmatze @squat @kakkoyun @bwplotka @krasi-georgiev

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -14,8 +14,6 @@ import (
 	"github.com/golang/snappy"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/prompb"
-
-	"github.com/openshift/telemeter/pkg/authorize"
 )
 
 const forwardTimeout = 5 * time.Second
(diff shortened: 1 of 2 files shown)


pkg/receive/handler.go
```diff
@@ -79,7 +77,7 @@ func (h *Handler) Receive(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	req = req.WithContext(ctx)
-	req.Header.Add("THANOS-TENANT", r.Context().Value(authorize.TenantKey).(string))
```
squat, line 82: maybe we should make this configurable via a flag so we don't have to make a new PR to telemeter every time the tenant ID changes. WDYT?

brancz: why would it change? it's kind of final ..

squat: when we run in different environments, or someone else runs this somewhere

metalmatze: Hopefully, this is replaced by OIDC soo anyway, no? If so, we could just leave it as is and then update to the proper new flow later.

brancz:

PR #333, opened 2020-05-05
telemeter-server: Make downstream tenant ID configurable

brancz:
Follow up from #332 

@squat @metalmatze @kakkoyun @krasi-georgiev @bwplotka


brancz:

Hopefully our last one for a while :)

PR #334, opened 2020-05-12
Bug 1813894: Stop using service ca from service account token

marun:
/cc @sttts @mfojtik @deads2k


squat:
now that we have switched our prod cluster to 4.x we can update this :tada: 
thanks @marun

PR #337, opened 2020-05-25
*: Eliminate global prometheus registry usage

kakkoyun:


metalmatze:
Nice. Generally looks like an amazing improvement. Is there anything specific I should review?

kakkoyun:
@metalmatze It'd be good have some feedback on the patterns or if you see anything odd? Other than that I just need to fix CI issues. All the checks work locally, even the benchmarks with using CodeReadyContainers. So I couldn't pinpoint the issues.

cmd/telemeter-server/main.go
```diff
@@ -350,6 +363,12 @@ func (o *Options) Run(ctx context.Context, externalListener, internalListener ne
 			internalListener.Close()
 		})
 	}
+
+	insMiddleware := server.NewNopInstrumentationMiddleware()
+	if reg != nil {
```
squat, line 368: we control the definition of reg, so why would it ever be nil?

pkg/forwarder/forwarder.go
```diff
@@ -91,23 +103,23 @@ type Worker struct {
...
 	lock        sync.Mutex
 	reconfigure chan struct{}
-
-	logger log.Logger
 }
 
 // New creates a new Worker based on the provided Config. If the Config contains invalid
 // values, then an error is returned.
-func New(cfg Config) (*Worker, error) {
+func New(metrics *Metrics, cfg Config) (*Worker, error) {
```
squat, line 110: why do we have to do this as another argument to the `New` func and not simply create the metrics inside of this creator func? That would simplify the life of the caller.

pkg/metricsclient/metricsclient.go
```diff
@@ -16,44 +16,51 @@ import (
...
+	logger  log.Logger
+	metrics *Metrics
+
+	client   *http.Client
+	maxBytes int64
+	timeout  time.Duration
 }
 
-func New(logger log.Logger, client *http.Client, maxBytes int64, timeout time.Duration, metricsName string) *Client {
+func New(logger log.Logger, metrics *Metrics, client *http.Client, maxBytes int64, timeout time.Duration) *Client {
```
squat, line 55: Same here, can we create the metrics directly inside of this creator func? And avoid having to expose another type "Metrics" and "NewMetrics"? That would simplify the package, the caller code etc. Also, it's really ambiguous what the `Metrics` type of the `metricsclient` package does.

PR #339, opened 2020-06-08
Pass oauth2.Transport into forward HTTP client

metalmatze:
This should now add the necessary auth tokens to every Telemeter forward request to the upstream /receive endpoint.

/cc @squat


cmd/telemeter-server/main.go
```diff
@@ -428,14 +428,43 @@ func (o *Options) Run() error {
...
+				Transport: telemeter_http.NewInstrumentedRoundTripper("forward",
+					&http.Transport{
+						MaxIdleConnsPerHost: 10,
+						IdleConnTimeout:     30 * time.Second,
+						DialContext:         (&net.Dialer{Timeout: 10 * time.Second}).DialContext,
+					},
+				),
+			}
+
+			if o.OIDCIssuer != "" {
```
squat, line 442: We already check for the existince of an OIDC issuer and set up an OIDC client and oauth client credentials on https://github.com/openshift/telemeter/blob/master/cmd/telemeter-server/main.go#L246. Can we simply reuse this to avoid code duplication?

metalmatze: Yes, let me check how to make the first one not as scoped up there so we can more easily reuse it.

cmd/telemeter-server/main.go
```diff
@@ -42,6 +37,10 @@ import (
 	"github.com/openshift/telemeter/pkg/metricfamily"
 	"github.com/openshift/telemeter/pkg/receive"
 	"github.com/openshift/telemeter/pkg/server"
+	"github.com/prometheus/client_golang/prometheus"
```
squat, line 40: I think this was done by mistake. These imports should be grouped with the non telemeter imports above

cmd/telemeter-server/main.go
```diff
@@ -242,30 +241,45 @@ func (o *Options) Run() error {
...
-			authorizeClient.Transport = &oauth2.Transport{
-				Base:   authorizeClient.Transport,
-				Source: cfg.TokenSource(ctx),
-			}
+		authorizeClient.Transport = &oauth2.Transport{
+			Base:   authorizeClient.Transport,
+			Source: cfg.TokenSource(ctx),
+		}
+		forwardClient.Transport = &oauth2.Transport{
+			Base:   authorizeClient.Transport,
```
squat, line 281: This should be `forwardClient.Transport`, othewise we are double wrapping a transport in OAuth2 roundtrippers

cmd/telemeter-server/main.go
```diff
@@ -242,30 +241,45 @@ func (o *Options) Run() error {
...
 
-			cfg := clientcredentials.Config{
-				ClientID:     o.ClientID,
-				ClientSecret: o.ClientSecret,
-				TokenURL:     provider.Endpoint().TokenURL,
-			}
+		ctx = context.WithValue(ctx, oauth2.HTTPClient,
+			&http.Client{
+				Timeout:   20 * time.Second,
+				Transport: telemeter_http.NewInstrumentedRoundTripper("oauth", authorizeClient.Transport),
```
squat, line 266: This should wrap the basic transport we created earlier `transport` rather than wrapping the instrumented authorize transport.

cmd/telemeter-server/main.go
```diff
@@ -242,30 +241,45 @@ func (o *Options) Run() error {
...
 
-		if o.OIDCIssuer != "" {
-			provider, err := oidc.NewProvider(ctx, o.OIDCIssuer)
-			if err != nil {
-				return fmt.Errorf("OIDC provider initialization failed: %v", err)
-			}
+	forwardClient := http.Client{
+		Timeout: 5 * time.Second,
+		Transport: telemeter_http.NewInstrumentedRoundTripper("forward",
+			&http.Transport{
```
squat, line 249: we should reuse the default transport we created earlier, simply named `transport`, for this since it has the same options set but most importantly because it respects the `verbose` setting

PR #320, opened 2020-03-24
Bug 1813221: jsonnet/telemeter: set cpu requests, unset limits

s-urbaniak:
/cc @openshift/openshift-team-monitoring

PR #346, opened 2020-07-15
Bug 1818806: jsonnet/telemeter: set cpu requests, unset limits

simonpasquier:
This is a backport of #320.

cc @openshift/openshift-team-monitoring

### DIFF
--- a/jsonnet/jsonnetfile.json
+++ b/jsonnet/jsonnetfile.json
@@ -1,7 +1,7 @@
 {
+  "version": 1,
   "dependencies": [
     {
-      "name": "telemeter",
       "source": {
         "local": {
           "directory": "telemeter"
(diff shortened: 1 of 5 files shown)


simonpasquier:
@bparees could you set `cherry-pick-approved` on this change? As Serg [wrote](https://github.com/openshift/telemeter/pull/320#issuecomment-604635871) in the PR for the `release-4.4` branch, this change is only touching the jsonnet code and the "real" work will happen in github.com/openshift/cluster-monitoring where we'll update the vendored version of the jsonnet library.

bparees:
> @bparees could you set cherry-pick-approved on this change?

I cannot, that's a patch manager responsibility(and i'm not patch manager this week).  You'll have to wait for the patch manager of the week to review+approve it(or explain why they did not/will not).  Should happen today.

simonpasquier:
Thanks for the quick and clear explanation!

PR #349, opened 2020-08-21
Bug 1873428: Fixed mem-leak and added proper e2e that uses prod code.

bwplotka:
Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>


cmd/telemeter-server/main.go
```diff
@@ -84,18 +82,33 @@ func main() {
...
+
+	var listen, listenInternal string
 	cmd := &cobra.Command{
 		Short:         "Aggregate federated metrics pushes",
 		Long:          desc,
 		SilenceErrors: true,
 		SilenceUsage:  true,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return opt.Run()
+			listener, err := net.Listen("tcp", listen)
```
squat, line 97: all these changes look ok but they are unrelated to the fix. let's merge them in another PR so that we can keep the changes small and easy to reason about

squat:
we need to rebase on the change in https://github.com/openshift/telemeter/pull/353 in order for the integration tests to pass

metalmatze:
Yes, I'm waiting on the #353 to pass. You think we should rebase already?

squat:
prow decided to restart some of the e2e tests because they were old. We can rebase on the one commit if we want or wait. Note, we need @bwplotka to prefix the PR title with `Bug 1873428: ` in order for Tide to accept this.

metalmatze:
Rebased to include the integration test fixes.

bwplotka (commented):
/lgtm

PR #352, opened 2020-08-25
Fix Telemeter integration-v2 tests

metalmatze:
Something is broken and I'll try to fix it in this PR. I'll get messy, but I'll try to squash it before merging. :upside_down_face:

### DIFF
--- a/test/integration-v2.sh
+++ b/test/integration-v2.sh
@@ -48,6 +48,7 @@ done
     --listen localhost:9103 \
     --listen-internal localhost:9104 \
     --forward-url=http://localhost:9105/api/v1/receive \
+    --memcached=localhost:11211 \
     -v
 ) &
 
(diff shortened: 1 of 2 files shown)


s-urbaniak:
thank you! :tada:

test/integration-v2.sh
```diff
@@ -48,6 +48,7 @@ done
     --listen localhost:9103 \
     --listen-internal localhost:9104 \
     --forward-url=http://localhost:9105/api/v1/receive \
+    --memcached=localhost:11211 \
```
squat: how did you add this line? this is already in the file: https://github.com/openshift/telemeter/blob/master/test/integration-v2.sh#L51

Are you working on the latest of the master branch?

squat: ah nvm, you removed it and added it back

metalmatze:
The actual fix is in #353.

PR #355, opened 2020-08-31
Bug 1874112: Do not overwrite 3xx and 4xx from Tollbooth to 500

metalmatze:
We really want to propagate the status codes we got from Tollbooth, so we don't alert on too many 5xx if we get 4xx from Tollbooth...


squat (commented):
/lgtm

bwplotka (approved):
/lgtm

PR #356, opened 2020-09-09
Bug 1877816: Provide synthetic recording rules for common cluster id -> key dimensional joins

smarterclayton:
In order to make estimation of cluster distribution easier, synthesize the following recording rules that remove the need for complex joins in other queries by _id:

* `version` - the current version of the cluster or empty string if the cluster did not report the metric
* `host_type` - an enumeration representing the most common types of hardware within a cluster by node count - metal, known virt types, cloud types (which may cover hardware)
* `provider` - an enumeration representing the cloud provider type when known
* `install_type` - an enumeration representing how the cluster was installed

Each of these attempt to classify all clusters reporting to telemetry:

* Clusters older than the metrics that started reporting these values
* Clusters with topologies that are missing subsets of metrics (hypershift clusters are not reporting control plane metrics)
* Clusters with different interpretation of these values over different releases

This also changes the `cluster_subscribed` metric to include these new dimensions to improve dimensionality reduction queries (show me upi aws clusters on version 4.4 that are subscribed)


s-urbaniak:
cc'ing @squat @metalmatze regarding the failing integration tests :thinking:

squat:
/test integration

There was a race between startup and sending metrics to we got 95% correctness instead of 100%. retesting

squat (approved):
/lgtm

PR #357, opened 2020-09-14
Bug 1879009: jsonnet/telemeter: Fix cluster_subscribed recording rule

smarterclayton:
Now that real data is being generated, fix the following problems
in the rules:

1. typo use of `provider` as the query instead of `id_provider`
2. fix multiple series generated for the same _id for id_version and
   id_install_type
3. Guard the recording rule for duplicate series in the future

### DIFF
--- a/jsonnet/telemeter/rules.libsonnet
+++ b/jsonnet/telemeter/rules.libsonnet
@@ -33,7 +33,7 @@
             {
               record: 'id_version_ebs_account_internal:cluster_subscribed',
               expr: |||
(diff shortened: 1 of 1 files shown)


metalmatze:
I might be worth thinking about improving readability by having these recording rules spread across multiple lines. An example can be found here: https://github.com/prometheus/prometheus/blob/master/documentation/prometheus-mixin/alerts.libsonnet#L172-L182

Nothing urgent, it's just slightly harder to grasp right now.

squat:
Let's fix this in prod now and when can make it pretty after

squat:
Semi urgent since the recording rules are kind of expensive right now and we don't want to unnecessarily abuse the ruler

metalmatze:
Yes, totally understandable to fix first for sure! :+1:

squat:
/title Bug 1879009: jsonnet/telemeter: Fix cluster_subscribed recording rule

squat:
/retitle Bug 1879009: jsonnet/telemeter: Fix cluster_subscribed recording rule

PR #358, opened 2020-09-15
Bug 1879965: jsonnet/telemeter: Use `host_type` in recording rule join

smarterclayton:
Joining on `primary_host_type` was incorrect and did not add the label.


squat (approved):
/lgtm

squat:
/retitle Bug 1879965: jsonnet/telemeter: Use `host_type` in recording rule join

PR #361, opened 2020-11-05
jsonnet/telemeter: Record which cloudpak a cluster appears to be

smarterclayton:
All known cloudpaks install at least one of the 'ibm-common-service'
operator or the 'ibm-licensing-operator'. To better track the known
cloudpaks report a metric that encodes the cloudpak_type of a cluster.
For now the only value reported is `unknown` to leave room to detect
type in the future.

A subsequent change will add this to the summary recording rule.


smarterclayton:
@squat this is just the new recording rule (tested directly), does not integrate it into the overall rule so we can verify it's working

squat:
ack, so that we don't prematurely break  before verifying.

/lgtm

bwplotka (approved):
/lgtm

smarterclayton:
This is still manual redeploy, right?

squat:
ack yes it is. I'll trigger a redeploy today :rocket:

PR #363, opened 2020-12-03
jsonnet/telemeter: Add a rule for normalizing ebs_account data

smarterclayton:
Combine the class and email_domain fields on subscription labels into
a smaller subset query. Add internal as it is used today for legacy
behavior, but leverage class to better discriminate between different
types of accounts (Internal, Customer, Partner, or Evaluation). The
resulting account_type should be used for classification of the
customer type.

Will be used for other joins and significantly reduces cardinality
of queries by avoiding subscription_labels


squat (approved):
/lgtm

smarterclayton:
Has this made it to stage or prod yet?

squat:
no, not yet; builds are failing due to the extra comma on this line: https://github.com/openshift/telemeter/pull/363/files#diff-be5c1148fb8fb67639520db2cee9afe1d431aad3afeeb44ed20b1ab85660a0e5R69

submitting a patch now

PR #366, opened 2021-01-11
Bug 1919360: Add ACM Managed cluster reporting rule

itdove:
https://github.com/open-cluster-management/backlog/issues/6365


simonpasquier:
/retitle  Bug 1919360: Add ACM Managed cluster reporting rule

PR #375, opened 2021-05-03
jsonnet/telemeter: Add a recording rule that captures network_type

smarterclayton:
Leverage knowledge about what resources are provided by a plugin
to categorize the usage of those plugins at a cluster _id level.
This is primarily accurate in late 4.6/4.7+ due to telemetry
including these resources. This categorized a significant chunk
of clusters although there are some edge cases noted.

### DIFF
--- a/jsonnet/telemeter/rules.libsonnet
+++ b/jsonnet/telemeter/rules.libsonnet
@@ -67,6 +67,26 @@
               |||,
             },
             {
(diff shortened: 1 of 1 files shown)


smarterclayton:
Tested manually on prod data set, no errors in this recording rule

squat (approved):
/lgtm

smarterclayton:
No bug needed because this is not in product.

/retest

squat:
yes :/ rolling this out and an update to the telemeter-server allow-list today :+1:

PR #379, opened 2021-06-16
Update dependency to generate UUID in benchmark

PhilipGough:
The existing library has issues with generating non-random UUID's
See https://github.com/satori/go.uuid/issues/73 for context.

Both libraries generate a v4 UUID, [RFC4412](https://datatracker.ietf.org/doc/html/rfc4122), so although there is no existing test to the package that was changed, it should be a safe to do so since it is tooling as opposed to hot path of the code.

~I  have also updated some details in the Makefile which pointed at an incorrect binary since `make build` builds `telemeter-benchmark` and not `benchmark`~


Makefile
```diff
@@ -26,7 +26,7 @@ JSONNET_LOCAL_OR_INSTALLED=$(if $(shell which jsonnet 2>/dev/null),$(shell which
 JSONNETFMT_LOCAL_OR_INSTALLED=$(if $(shell which jsonnetfmt 2>/dev/null),$(shell which jsonnetfmt 2>/dev/null),$(JSONNETFMT_BIN))
 JB_BIN=$(BIN_DIR)/jb
 JSONNET_SRC=$(shell find ./jsonnet -path ./jsonnet/vendor -prune -false -o -type f \( -name "*.jsonnet" -o -name "*.libsonnet" \))
-BENCHMARK_RESULTS=$(shell find ./benchmark -type f -name '*.json')
+BENCHMARK_RESULTS=$(shell find ./telemeter-benchmark -type f -name '*.json')
```
squat, line 29: I don't think this is right. The telemeter-benchmark binary is not what we are looking for here, but rather we are looking for JSON files in a directory called `benchmark`

squat (changes requested):
I think we need to revert some unrelated changes but otherwise looks good to me :+1:

Makefile
PhilipGough: Got it, my bad, let me revert the second commit

PR #383, opened 2021-07-05
remove AppSRE from owners

maorfr:
When things started out and we had just a few services to run, it made sense that we attempt to be approvers on each of the repositories.
As we manage dozens of services with hundreds of repos, this approach no longer scales and we had decided to not proceed with it. Here is the MR to our onboarding docs which removes this section: https://gitlab.cee.redhat.com/service/app-interface/-/commit/5dd6d564a822ee05eedb7622ea672ea6cc37ebf7

As such, I am submitting this PR to remove all of AppSRE from being approvers.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -6,19 +6,10 @@ reviewers:
 - squat
 - s-urbaniak
 - smarterclayton
-- aditya-konarde
-- jfchevrette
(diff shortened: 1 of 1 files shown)


squat (approved):
/lgtm

PR #385, opened 2021-07-08
add app.kubernetes.io/part-of label to telemeter client

spaparaju:
To telemeter-client, add `app.kubernetes.io/part-of` label to [refer  at telemeter-client deployed by CMO](https://github.com/openshift/cluster-monitoring-operator/pull/1205/#discussion_r660375686) .

### DIFF
--- a/jsonnet/telemeter/client.libsonnet
+++ b/jsonnet/telemeter/client.libsonnet
@@ -6,5 +6,9 @@
     telemeterClient+: {
       matchRules+: [],
     },
+    commonLabels+:: {
+      'app.kubernetes.io/component': 'telemetry-metrics-collector',
+      'app.kubernetes.io/name': 'telemeter-client',
+    },
   },
 }
--- a/jsonnet/telemeter/client/kubernetes.libsonnet
+++ b/jsonnet/telemeter/client/kubernetes.libsonnet
@@ -92,7 +92,7 @@ local securePort = 8443;
       local containerVolumeMount = container.volumeMountsType;
       local containerEnv = container.envType;
 
-      local podLabels = { 'k8s-app': 'telemeter-client' };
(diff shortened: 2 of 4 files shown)


jsonnet/telemeter/client.libsonnet
```diff
@@ -6,5 +6,10 @@
     telemeterClient+: {
       matchRules+: [],
     },
+
+    commonLabels+:: {
+      'app.kubernetes.io/part-of': 'openshift-monitoring',
```
squat, line 11: This field/object doesn't seem to be used anywhere in the kubernetes.libsonnet code. I think you need to either manually add it here [0] or use the `commonLabels` field in that file somewhere. 

[0[ https://github.com/openshift/telemeter/blob/master/jsonnet/telemeter/client/kubernetes.libsonnet#L95

spaparaju: 👍

squat (approved):
/lgtm

jsonnet/telemeter/client/kubernetes.libsonnet
```diff
@@ -92,7 +92,7 @@ local securePort = 8443;
       local containerVolumeMount = container.volumeMountsType;
       local containerEnv = container.envType;
 
-      local podLabels = { 'k8s-app': 'telemeter-client' };
+      local podLabels = $._config.commonLabels + { 'k8s-app': 'telemeter-client' };
```
dgrisonnet, line 95: Now that we are using the recommended Kubernetes labels, we can remove k8s-app.
```suggestion
      local podLabels = $._config.commonLabels;
```

spaparaju: 👍

PR #392, opened 2021-10-15
MON-1964: Increase limit receive bytes default to 128kB

jan--f:
There is a second commit that adds logging to `LimitBodySize`. Happy to discuss this one, just thought it would be handy to be able to get that info.

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -101,10 +101,12 @@ func (h *Handler) Receive(w http.ResponseWriter, r *http.Request) {
 }
 
 // LimitBodySize is a middleware that check that the request body is not bigger than the limit
-func LimitBodySize(limit int64, next http.Handler) http.HandlerFunc {
+func LimitBodySize(logger log.Logger, limit int64, next http.Handler) http.HandlerFunc {
+	logger = log.With(logger, "component", "receive", "middleware", "LimitBodySize")
 	return func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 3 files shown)


pkg/receive/handler.go
```diff
@@ -114,6 +116,7 @@ func LimitBodySize(limit int64, next http.Handler) http.HandlerFunc {
 		r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
 
 		if len(body) >= int(limit) {
+			level.Debug(logger).Log("msg", "request is too big", "req_size", len(body))
```
matej-g, line 119: :clap:

pkg/receive/handler.go
```diff
@@ -19,7 +19,7 @@ import (
 const forwardTimeout = 5 * time.Second
 
 // DefaultRequestLimit is the size limit of a request body coming in
-const DefaultRequestLimit = 15 * 1024 // based on historic Prometheus data with 6KB at most
+const DefaultRequestLimit = 128 * 1024
```
matej-g, line 22: Shall we make the request limit configurable via flag? Or you'd like to change the default limit first? I'm thinking it would be more future-proof, in case we find out this limit does not work and we need to change it again.

simonpasquier: https://github.com/openshift/telemeter/pull/391 was about adding the flag or am I mistaken?

jan--f: Yes, in #391 we add the cli arg with `DefaultRequestLimit` as the default value.

PR #413, opened 2022-06-30
Updates version of golangci-lint to support GO1.18

JoaoBraveCoding:
ART builds were breaking due to golangci-lint


JoaoBraveCoding:
/retitle Bug 2104549: Updates version of golangci-lint to support GO1.18

PR #418, opened 2022-07-28
Add rules for cluster CPU-hours and Instance-hours

kahowell:
These are intended to help simplify PromQL used by subscription watch,
as well as make it more visible to others.

Additionally, the encapsulation allows us to tweak the definition of
CPU-hours or instance-hours if better underlying metrics are made
available.

Note this is untested, but I'm happy to assist with testing, but I lack context/fixtures.

### DIFF
--- a/jsonnet/telemeter/rules.libsonnet
+++ b/jsonnet/telemeter/rules.libsonnet
@@ -142,10 +142,9 @@
             {
               // OpenShift Cluster Instance-hours for the last hour
               // The divisor of scalar(count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
-              // max(...) by (_id) is used ensure a single datapoint per cluster ID
               record: 'cluster:usage:workload:capacity_physical_instance_hours',
               expr: |||
-                max(sum_over_time(group(cluster:usage:workload:capacity_physical_cpu_cores:max:5m) by (_id)[1h:5m]) / scalar(count_over_time(vector(1)[1h:5m]))) by (_id)
+                max by(_id) (count_over_time(cluster:usage:workload:capacity_physical_cpu_cores:max:5m[1h:5m])) / scalar(count_over_time(vector(1)[1h:5m]))
               |||,
             },
           ],
--- a/jsonnet/telemeter/rules.libsonnet
+++ b/jsonnet/telemeter/rules.libsonnet
@@ -131,16 +131,16 @@
               |||,
             },
             {
-              // OpenShift Cluster CPU-hours for the last hour
+              // OpenShift Cluster CPU-hours for the last hour.
               // The divisor of scalar(count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
-              // max(...) by (_id) is used ensure a single datapoint per cluster ID
+              // max(...) by (_id) is used ensure a single datapoint per cluster ID.
               record: 'cluster:usage:workload:capacity_physical_cpu_hours',
               expr: |||
                 max by(_id) (sum_over_time(cluster:usage:workload:capacity_physical_cpu_cores:max:5m[1h:5m]) / scalar(count_over_time(vector(1)[1h:5m])))
               |||,
             },
             {
-              // OpenShift Cluster Instance-hours for the last hour
+              // OpenShift Cluster Instance-hours for the last hour.
               // The divisor of scalar(count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
               record: 'cluster:usage:workload:capacity_physical_instance_hours',
               expr: |||
--- a/jsonnet/telemeter/rules.libsonnet
+++ b/jsonnet/telemeter/rules.libsonnet
@@ -132,20 +132,20 @@
             },
             {
               // OpenShift Cluster CPU-hours for the last hour
-              // The divisor of count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
+              // The divisor of scalar(count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
               // max(...) by (_id) is used ensure a single datapoint per cluster ID
               record: 'cluster:usage:workload:capacity_physical_cpu_hours',
               expr: |||
-                max(sum_over_time(cluster:usage:workload:capacity_physical_cpu_cores:max:5m[1h:5m]) / count_over_time(vector(1)[1h:5m])) by (_id)
+                max(sum_over_time(cluster:usage:workload:capacity_physical_cpu_cores:max:5m[1h:5m]) / scalar(count_over_time(vector(1)[1h:5m]))) by (_id)
               |||,
             },
             {
               // OpenShift Cluster Instance-hours for the last hour
-              // The divisor of count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
+              // The divisor of scalar(count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
               // max(...) by (_id) is used ensure a single datapoint per cluster ID
               record: 'cluster:usage:workload:capacity_physical_instance_hours',
               expr: |||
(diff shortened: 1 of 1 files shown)


jsonnet/telemeter/rules.libsonnet
```diff
@@ -130,6 +130,24 @@
                 topk(500, sum (acm_managed_cluster_info) by (managed_cluster_id, cloud, created_via, endpoint, instance, job, namespace, pod, service, vendor, version))
               |||,
             },
+            {
+              // OpenShift Cluster CPU-hours for the last hour
+              // The divisor of count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
+              // max(...) by (_id) is used ensure a single datapoint per cluster ID
+              record: 'cluster:usage:workload:capacity_physical_cpu_hours',
+              expr: |||
+                max(sum_over_time(cluster:usage:workload:capacity_physical_cpu_cores:max:5m[1h:5m]) / count_over_time(vector(1)[1h:5m])) by (_id)
```
simonpasquier, line 139: In the current form, the expression would return no data because the right-hand side has no `_id` label to match with the left-hand side (vector() returns a scalar as a vector with no label).
`count_over_time(vector(1)[1h:5m])` is always going to return 12 anyway. Hence it could be replaced by `12` (if it's really what you wanted).

jsonnet/telemeter/rules.libsonnet
```diff
@@ -130,6 +130,24 @@
...
+              |||,
+            },
+            {
+              // OpenShift Cluster Instance-hours for the last hour
+              // The divisor of count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
+              // max(...) by (_id) is used ensure a single datapoint per cluster ID
+              record: 'cluster:usage:workload:capacity_physical_instance_hours',
+              expr: |||
+                max(sum_over_time(group(cluster:usage:workload:capacity_physical_cpu_cores:max:5m) by (_id)[1h:5m]) / count_over_time(vector(1)[1h:5m])) by (_id)
```
simonpasquier, line 148: `sum_over_time(group(cluster:usage:workload:capacity_physical_cpu_cores:max:5m) by (_id))[1h:5m])` should be the same as `max by(_id) (count_over_time(cluster:usage:workload:capacity_physical_cpu_cores:max:5m[1h:5m]))`?

jsonnet/telemeter/rules.libsonnet
kahowell: Ah, I was missing a `scalar` call. I PoC'd this change and then forgot to include `scalar` when I transcribed it.

jsonnet/telemeter/rules.libsonnet
```diff
@@ -130,6 +130,23 @@
                 topk(500, sum (acm_managed_cluster_info) by (managed_cluster_id, cloud, created_via, endpoint, instance, job, namespace, pod, service, vendor, version))
               |||,
             },
+            {
+              // OpenShift Cluster CPU-hours for the last hour
+              // The divisor of scalar(count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
+              // max(...) by (_id) is used ensure a single datapoint per cluster ID
+              record: 'cluster:usage:workload:capacity_physical_cpu_hours',
+              expr: |||
+                max(sum_over_time(cluster:usage:workload:capacity_physical_cpu_cores:max:5m[1h:5m]) / scalar(count_over_time(vector(1)[1h:5m]))) by (_id)
```
simonpasquier, line 139: scalar(count_over_time(vector(1)[1h:5m])) will always return 12 so you could simplify the expression.

Personal preference: "max by(...) (...)" reads better than "max(...) by(...)"

```suggestion
                max by (_id) (sum_over_time(cluster:usage:workload:capacity_physical_cpu_cores:max:5m[1h:5m]) / 12)
```

jsonnet/telemeter/rules.libsonnet
kahowell: As to the 12 thing, the reason I thought it might be wise to use `scalar(count_over_time(vector(1)[1h:5m]))` instead was that I noticed different answers for `scalar(count_over_time(vector(1)[1h:5m]))` depending on step and timestamp passed. For example:

<details>
<summary>Expand for screenshot</summary>

image
</details>
In practice, we noticed that when using step=3600 and aligned to the top of the hour, we always get a value of `13`:

<details>
<summary>Expand for screenshot</summary>

image
</details>

I figured this was something to do with Prometheus doing sampling (actually, I'd love a more specific/accurate explanation, if you have one). Thus it seemed unsafe to simply use 13, since, if I understand correctly the recording rule doesn't necessarily run at the top of the hour. If you believe 13 is fine (or 12), though, I am more than happy to hardcode.

jsonnet/telemeter/rules.libsonnet
kahowell: Done, much thanks for the suggestion!

jsonnet/telemeter/rules.libsonnet
kahowell: I actually like `max by` better too, now that I've seen it :-). I went ahead and made that change, I am looking forward to your thoughts on `12` vs. `13` vs. `scalar(count_over_time(vector(1)[1h:5m])))`. See above thread please.

jsonnet/telemeter/rules.libsonnet
simonpasquier: hmm this looks like a Thanos artifact. With a vanilla Prometheus, I can't reproduce...

image

kahowell: @simonpasquier given the above, how should we proceed?
* hardcode `12`?
* hardcode `13`?
* use `scalar(count_over_time(vector(1)[1h:5m]))`?

simonpasquier: `scalar(count_over_time(vector(1)[1h:5m]))` is probably going to return the "correct" result but I'd be eager to hear from @bwplotka if it's something that he's aware of.

barnabycourt: @bwplotka following up on the question that went your August 2. Does this look ok to you?

accorvin:
@anishasthana I think you have a lot of experience with the rules we're using like this in RHODS - can you take a look at this too and give your opinion on whether it could satisfy RHODS use cases from your perspective?

anishasthana:
After talking to Jeff and Kevin, I think these rules would satisfy RHODS use cases.

jsonnet/telemeter/rules.libsonnet
```diff
@@ -130,6 +130,23 @@
                 topk(500, sum (acm_managed_cluster_info) by (managed_cluster_id, cloud, created_via, endpoint, instance, job, namespace, pod, service, vendor, version))
               |||,
             },
+            {
+              // OpenShift Cluster CPU-hours for the last hour
+              // The divisor of scalar(count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
+              // max(...) by (_id) is used ensure a single datapoint per cluster ID
```
bwplotka, line 136:
```suggestion
              // OpenShift Cluster CPU-hours for the last hour.
              // The divisor of scalar(count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
              // max(...) by (_id) is used ensure a single datapoint per cluster ID.
```

jsonnet/telemeter/rules.libsonnet
```diff
@@ -130,6 +130,23 @@
...
+              record: 'cluster:usage:workload:capacity_physical_cpu_hours',
+              expr: |||
+                max by(_id) (sum_over_time(cluster:usage:workload:capacity_physical_cpu_cores:max:5m[1h:5m]) / scalar(count_over_time(vector(1)[1h:5m])))
+              |||,
+            },
+            {
+              // OpenShift Cluster Instance-hours for the last hour
```
bwplotka, line 143:
```suggestion
              // OpenShift Cluster Instance-hours for the last hour.
```

bwplotka (approved):
LGTM!

moadz (approved):

anishasthana:
Looks like we still need a lgtm on the PR.

PR #426, opened 2022-08-10
pkg/receive/handler: Use OIDC client

saswatamcode:
This PR makes sure that /metrics/v1/receive handler uses the OIDC forwardClient to send requests to Observatorium API instead.

### DIFF
--- a/cmd/telemeter-server/main.go
+++ b/cmd/telemeter-server/main.go
@@ -512,6 +512,7 @@ func (o *Options) Run(ctx context.Context, externalListener, internalListener ne
 		// v2 routes
 		{
 			v2AuthorizeClient := authorizeClient
+			v2ForwardClient := forwardClient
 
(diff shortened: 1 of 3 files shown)


bill3tt:
@saswatamcode have you been able to do any validation / testing of this fix? Not a blocker, this path is unused and buggy as is so we can't do much worse than we are doing right now.

bill3tt (approved):

saswatamcode:
@bill3tt so this path does work locally, using latest Prometheus + Thanos Receive, based on some local testing I did: https://github.com/saswatamcode/telemeter-prom-receive/blob/main/demo.sh.
However to actually test this with in-cluster Prometheus -> Telemeter Server -> Observatorium API, I need to deploy

simonpasquier:
/lgtm
with blessing from @bill3tt which isn't listed under the proper handle in the OWNERS file

PR #427, opened 2022-08-13
pkg/receive/handler: Add middleware to respect whitelist + elide

saswatamcode:
This PR adds a `TransformWriteRequest` middlerware to /metrics/v1/receive, to allow applying whitelist rules and elide labels to remote write requests.

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"time"
@@ -128,13 +129,13 @@ func (h *Handler) Receive(w http.ResponseWriter, r *http.Request) {
 func LimitBodySize(logger log.Logger, limit int64, next http.Handler) http.HandlerFunc {
 	logger = log.With(logger, "component", "receive", "middleware", "LimitBodySize")
 	return func(w http.ResponseWriter, r *http.Request) {
+		defer r.Body.Close()
 		body, err := ioutil.ReadAll(r.Body)
 		if err != nil {
 			level.Error(logger).Log("msg", "failed to read body", "err", err)
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
-		defer r.Body.Close()
 
 		// Set body to this buffer for other handlers to read
 		r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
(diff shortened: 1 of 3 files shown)


pkg/receive/handler.go
```diff
@@ -193,3 +212,80 @@ func ValidateLabels(logger log.Logger, next http.Handler, labels ...string) http
 		next.ServeHTTP(w, r)
 	}
 }
+
+func (h *Handler) TransformWriteRequest(logger log.Logger, next http.Handler) http.HandlerFunc {
+	logger = log.With(logger, "component", "receive", "middleware", "transformWriteRequest")
+	return func(w http.ResponseWriter, r *http.Request) {
+		body, err := ioutil.ReadAll(r.Body)
```
juanrh, line 219: This seems to be deprecated  https://pkg.go.dev/io/ioutil#ReadAll

> Deprecated: As of Go 1.16, this function simply calls io.ReadAll.

pkg/receive/handler.go
```diff
@@ -193,3 +212,80 @@ func ValidateLabels(logger log.Logger, next http.Handler, labels ...string) http
...
+		if err != nil {
+			level.Error(logger).Log("msg", "failed to read body", "err", err)
+			http.Error(w, "failed to read body", http.StatusInternalServerError)
+			return
+		}
+		defer r.Body.Close()
```
juanrh, line 225: Shouldn't this be moved before the `if`, to ensure we always close even on read error?

pkg/receive/handler.go
```diff
@@ -193,3 +212,80 @@ func ValidateLabels(logger log.Logger, next http.Handler, labels ...string) http
 		next.ServeHTTP(w, r)
 	}
 }
+
+func (h *Handler) TransformWriteRequest(logger log.Logger, next http.Handler) http.HandlerFunc {
+	logger = log.With(logger, "component", "receive", "middleware", "transformWriteRequest")
```
juanrh, line 217: `h.logger` uses `"component", "receive/handler"`, shouldn't this be the same? Another option is using `logger 
 = log.With(h.logger,  "middleware", "transformWriteRequest")`, what do you think?

juanrh (commented):
I don't have much context on this project, so I'm just leaving some general comments, and deferring a more proper review to someone else with more context

pkg/receive/handler.go
saswatamcode: Going with second option!

pkg/receive/handler.go
saswatamcode: Yeah, but the project was on Go 1.13, so was confused about this. Have changed.

pkg/receive/handler.go
saswatamcode: Yeah, I copied this from the `LimitBodySize` middleware. Have replaced in all places!

pkg/receive/handler.go
```diff
@@ -193,3 +213,80 @@ func ValidateLabels(logger log.Logger, next http.Handler, labels ...string) http
 		next.ServeHTTP(w, r)
 	}
 }
+
+func (h *Handler) TransformWriteRequest(logger log.Logger, next http.Handler) http.HandlerFunc {
```
douglascamata, line 217: I am only a tiny bit worried about the fact that we're doing the whole process of `snappy decode -> unmarshall -> filter -> marshall -> snappy encode` after the `ValidateLabels` middleware does the exact same thing. Isn't it too much extra work considering the scale that this service runs at? Would it be possible to avoid the extra work?

saswatamcode: Yeah, for now, there isn't a whole lot of usage on this endpoint, but would be good to improve and iterate on it! :)

PR #428, opened 2022-08-16
*:Update project with latest tools/dependencies

saswatamcode:
This PR updates several dependencies which were used by Telemeter to latest, and bumps the project to Go 1.17.

This also removes the `tools` dir, and replaces it with bingo. 

Diff is large due to vendoring.


saswatamcode:
I will follow up with another PR to convert the bash-based e2e tests to Go, not doing that here diff would be really large. This PR is ready for review. 🙂

PR #429, opened 2022-08-25
Forward: Add logic to sanitize labels

matej-g:
Signed-off-by: Matej Gera <matejgera@gmail.com>

Since Thanos Receiver from version `0.28.0` will require the labels to be valid (ordered, no duplicates and no empty labels), we need to ensure whatever we receive from client will contain valid labels.

Added tests to the conversion method to test that sanitizing labels works.


bwplotka (approved):
Amazing, thanks!

PhilipGough (commented):
/lgtm thanks

PR #431, opened 2022-09-26
Adds TemeterClientJobFailing alert

JoaoBraveCoding:
Issue: https://issues.redhat.com/browse/MON-2727

Problem: currently if the telemter client starts failing to push metrics to telemeter server, there is no alert to notify admins

Solution: this PR adds an alert to telemeter that leverages the federate_requests_failed_total to notify admins if this counter keeps increasing during 1h.


jsonnet/telemeter/client/kubernetes.libsonnet
```diff
@@ -244,6 +244,22 @@ local securePort = 8443;
                 record: 'cluster:telemetry_selected_series:count',
                 expr: 'max(federate_samples - federate_filtered_samples)',
               },
+              {
+                expr: ||| 
+                  sum by (job,namespace,pod) (
+                    rate(federate_requests_failed_total{job="telemeter-client"}[5m])
```
wking, line 250: I forget exactly how often the client attempts submissions, but I thought it was on the order of ~minutes.  If so, `[5m]` may only capture an attempt or two, which makes it hard for me to reason about `rate`.  Can we stretch this out to ensure we cover ~5 attempts or some such so we don't have to worry as much about the granularity of attempts?

jsonnet/telemeter/client/kubernetes.libsonnet
```diff
@@ -244,6 +244,22 @@ local securePort = 8443;
...
+                  ) > 0
+                |||,
+                labels: {
+                  severity: 'warning',
+                },
+                annotation: {
+                  description: 'Telemeter client in {{ $labels.namespace }}/{{ $labels.pod }} is failing to push metrics to telemeter server',
```
wking, line 257: Runbook URI or inline next-step suggestions?  Something that points folks at the relevent Events and/or logs would probably be the minimum needed to diagnose and fix whatever is going on.

jan--f: This might be better suited to add downstream. We have openshift/runbooks and could probably outline more context to possible root causes, which don't apply to upstream necessarily.

jsonnet/telemeter/client/kubernetes.libsonnet
simonpasquier: telemeter client sends data every 4m30s.

jsonnet/telemeter/client/kubernetes.libsonnet
```diff
@@ -244,6 +244,22 @@ local securePort = 8443;
                 record: 'cluster:telemetry_selected_series:count',
                 expr: 'max(federate_samples - federate_filtered_samples)',
               },
+              {
+                expr: ||| 
+                  sum by (job,namespace,pod) (
+                    rate(federate_requests_failed_total{job="telemeter-client"}[5m])
+                  ) > 0
```
simonpasquier, line 251: Alerting on the ratio of failed requests is usually more convenient and I'd go with a conservative threshold (e.g. at least 20%) since the main use case is to report when telemetry push fails repeatedly.

```suggestion
                  sum by (namespace) (
                    rate(federate_requests_failed_total{job="telemeter-client"}[15m])
                  ) /
                  sum by (namespace) (
                    rate(federate_requests_total{job="telemeter-client"}[15m])
                  ) > 0.2
```

jsonnet/telemeter/client/kubernetes.libsonnet
simonpasquier: I wouldn't mind to add more details here since it's openshift/telemeter isn't really upstream anyway.

JoaoBraveCoding: Added a command to get the logs and a couple of helpful pointers, let me know if it's enough

jsonnet/telemeter/client/kubernetes.libsonnet
JoaoBraveCoding: Ahhh yes! I was planning on extending the range to 15m but I would never have thought about using a ratio, indeed it makes way more sense! I'll try to remember next time!

jsonnet/telemeter/client/kubernetes.libsonnet
```diff
@@ -244,6 +244,25 @@ local securePort = 8443;
...
+                  ) > 0.2
+                |||,
+                labels: {
+                  severity: 'warning',
+                },
+                annotation: {
+                  description: 'Telemeter client in {{ $labels.namespace }}/{{ $labels.pod }} is failing to push metrics to telemeter server.\nTo view the logs of telemeter-client execute the command: "oc logs -n openshift-monitoring deployment.apps/telemeter-client -c telemeter-client"\nIf telemeter-client fails to authenticate with telemeter-server make sure the gloabl pull secret is up to date, see https://docs.openshift.com/container-platform/4.11/openshift_images/managing_images/using-image-pull-secrets.html#images-update-global-pull-secret_using-image-pull-secrets\nTelemeter-client workflow is described in https://github.com/openshift/telemeter#upload-endpoint-receive-metrics-in-client_modelmetricfamily-format-from-telemeter-client-currently-used-by-cmo',
+                  summary: 'Telemeter client is failing to push metrics',
+                },
+                alert: 'TelemeterClientJobFailing',
```
simonpasquier, line 263:
```suggestion
                alert: 'TelemeterClientFailures',
```

jsonnet/telemeter/client/kubernetes.libsonnet
```diff
@@ -244,6 +244,25 @@ local securePort = 8443;
...
+                  ) /
+                  sum by (namespace) (
+                    rate(federate_requests_total{job="telemeter-client"}[15m])
+                  ) > 0.2
+                |||,
+                labels: {
+                  severity: 'warning',
+                },
+                annotation: {
+                  description: 'Telemeter client in {{ $labels.namespace }}/{{ $labels.pod }} is failing to push metrics to telemeter server.\nTo view the logs of telemeter-client execute the command: "oc logs -n openshift-monitoring deployment.apps/telemeter-client -c telemeter-client"\nIf telemeter-client fails to authenticate with telemeter-server make sure the gloabl pull secret is up to date, see https://docs.openshift.com/container-platform/4.11/openshift_images/managing_images/using-image-pull-secrets.html#images-update-global-pull-secret_using-image-pull-secrets\nTelemeter-client workflow is described in https://github.com/openshift/telemeter#upload-endpoint-receive-metrics-in-client_modelmetricfamily-format-from-telemeter-client-currently-used-by-cmo',
```
simonpasquier, line 260:
```suggestion
                  description: 'The telemeter client in namespace {{ $labels.namespace }} fails {{ $value | humanize }} of the requests to the telemeter service.\nCheck the logs of the telemeter-client pod with the following command:\noc logs -n openshift-monitoring deployment.apps/telemeter-client -c telemeter-client\nIf the telemeter client fails to authenticate with the telemeter service, make sure that the global pull secret is up to date, see https://docs.openshift.com/container-platform/latest/openshift_images/managing_images/using-image-pull-secrets.html#images-update-global-pull-secret_using-image-pull-secrets for more details.',
```

jsonnet/telemeter/client/kubernetes.libsonnet
```diff
@@ -244,6 +244,25 @@ local securePort = 8443;
...
+                  summary: 'Telemeter client is failing to push metrics',
```
simonpasquier, line 261:
```suggestion
                  summary: 'Telemeter client fails to send metrics',
```

simonpasquier (commented):
a few nits.

jsonnet/telemeter/client/kubernetes.libsonnet
```diff
@@ -244,6 +244,25 @@ local securePort = 8443;
                 record: 'cluster:telemetry_selected_series:count',
                 expr: 'max(federate_samples - federate_filtered_samples)',
               },
+              {
+                expr: |||
+                  sum by (namespace) (
```
juanrh, line 249: Why `by (namespace)`? The telemeter client is running in a single namespace, as mentioned in `annotation.description` below. Also, in github.com/openshift/telemeter/pkg/forwarder/forwarder.go I see no labels added, so the namespace should be added as a target label, which will always be `openshift-monitoring` because this is running just in that namespace

jsonnet/telemeter/client/kubernetes.libsonnet
```diff
@@ -244,6 +244,25 @@ local securePort = 8443;
...
+                    rate(federate_requests_failed_total{job="telemeter-client"}[15m])
+                  ) /
+                  sum by (namespace) (
+                    rate(federate_requests_total{job="telemeter-client"}[15m])
+                  ) > 0.2
```
juanrh, line 254: Why 20% as the threshold? Checking similar alarms with `ag -G '\.*' 'failed_total' jsonnet/` I see PrometheusRemoteStorageFailures, AlertmanagerClusterFailedToSendAlerts and AlertmanagerFailedToSendAlerts with 1% threshold. Arguably, those alarms are more critical, but anyway 20% looks quite high, what is the rationale behind that value?

manifests/client/prometheusRule.yaml
```diff
@@ -9,3 +9,21 @@ spec:
     rules:
     - expr: max(federate_samples - federate_filtered_samples)
       record: cluster:telemetry_selected_series:count
+    - alert: TelemeterClientFailures
+      annotation:
+        description: |-
+          The telemeter client in namespace {{ $labels.namespace }} fails {{ $value | humanize }} of the requests to the telemeter service.
+          Check the logs of the telemeter-client pod with the following command:
+          oc logs -n openshift-monitoring deployment.apps/telemeter-client -c telemeter-client
+          If the telemeter client fails to authenticate with the telemeter service, make sure that the global pull secret is up to date, see https://docs.openshift.com/container-platform/latest/openshift_images/managing_images/using-image-pull-secrets.html#images-update-global-pull-secret_using-image-pull-secrets for more details.
```
juanrh, line 18: These explanations are quite specific, and there is a risk that the mention of a specific namespace and docs URL would get out of date in future versions. I see no other docs URL searching for `ag -G '\.*' docs.openshift.com jsonnet/`. Maybe it would be best to have a more generic message that is more maintainable, like for example [this previous one](https://github.com/openshift/cluster-monitoring-operator/blob/93df2f7e3e95efc232f5ed1de9e5f607ce619823/assets/prometheus-k8s/prometheus-rule.yaml#L156-L173), and also mention the percentage of failures that we use as threshold

juanrh (changes requested):

jsonnet/telemeter/client/kubernetes.libsonnet
juanrh: Nevermind, I saw Simon's previous comment

jsonnet/telemeter/client/kubernetes.libsonnet
simonpasquier: as stated in https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md, it's a good practice to have a namespace label attached to the alert as it enables better alert routing (e.g. all alerts from openshift-monitoring should go to the team responsible for the monitoring components). The namespace could be hard-coded in the fixed alert's labels or (preferred) come from the metric itself.

juanrh:
Thanks a lot for the clarifications @simonpasquier 
lgtm, but leaving a chance for others to comment

simonpasquier (commented):
/lgtm

PR #440, opened 2023-01-05
pkg/receive: Sanitize metric labels

saswatamcode:
Similar to https://github.com/openshift/telemeter/pull/429, adding label sanitization to the Telemeter server receive path,
as we need to ensure this for Thanos Receive with versions > `v0.28`.


PhilipGough (approved):
/lgtm

PR #443, opened 2023-01-11
Add metrics for telemeter receive path

saswatamcode:
This PR adds a couple of new counters to accurately measure any request failures due to body size limit and request label validations.

This also refactors handler to ensure we decode and encode remote-write protobufs only once for transforming/validating remote write requests.


danielmellado:
Overall codewise lgtm, let's wait for CI, thanks!

PhilipGough (approved):
/lgtm

PR #444, opened 2023-01-11
Bump snappy version and add debug logs

saswatamcode:
Bumps snappy version to latest and adds some debug logs to verify empty remote write requests

cc: @PhilipGough


PhilipGough (approved):
/lgtm

PR #449, opened 2023-02-03
server: Ensure logging level is flag is respected

PhilipGough:
The logging level was never respected since the flag value only become available within `Run`.

This change sets a levelled logger based on the value of the flag as expected!

Before and after screenshots of the debugger below.

Screenshot 2023-02-03 at 16 15 26
Screenshot 2023-02-03 at 16 19 33


douglascamata (approved):
LGTM 🚀

saswatamcode (approved):
/lgtm

PR #455, opened 2023-03-28
MON-3230: Add TLS auth to telemeter-client

marioferh:
## Description

Add TLS auth to forwarder in order to authorize telemeter against rbac Prometheus proxy. Change FROM to 9092 proxy port.

This PR works together with this CMO: https://github.com/openshift/cluster-monitoring-operator/pull/1904

### DIFF
--- a/cmd/telemeter-client/main.go
+++ b/cmd/telemeter-client/main.go
@@ -48,6 +48,11 @@ func main() {
 
 	cmd.Flags().StringVar(&opt.Listen, "listen", opt.Listen, "A host:port to listen on for health and metrics.")
 	cmd.Flags().StringVar(&opt.From, "from", opt.From, "The Prometheus server to federate from.")
+
+	// TLS flags
+	cmd.Flags().StringVar(&opt.TLSCertFile, "tls-cert-file", "", "File containing the default x509 Certificate for HTTPS. (CA cert, if any, concatenated after server cert)")
+	cmd.Flags().StringVar(&opt.TLSKey, "tls-private-key-file", "", "File containing the default x509 private key matching --tls-cert-file.")
+
 	cmd.Flags().StringVar(&opt.FromToken, "from-token", opt.FromToken, "A bearer token to use when authenticating to the source Prometheus server.")
 	cmd.Flags().StringVar(&opt.FromCAFile, "from-ca-file", opt.FromCAFile, "A file containing the CA certificate to use to verify the --from URL in addition to the system roots certificates.")
 	cmd.Flags().StringVar(&opt.FromTokenFile, "from-token-file", opt.FromTokenFile, "A file containing a bearer token to use when authenticating to the source Prometheus server.")
@@ -109,6 +114,8 @@ type Options struct {
 
 	From          string
 	To            string
+	TLSCertFile   string
+	TLSKey        string
 	ToUpload      string
 	ToAuthorize   string
 	FromCAFile    string
@@ -269,6 +276,8 @@ func (o *Options) Run() error {
 		FromTokenFile: o.FromTokenFile,
 		ToTokenFile:   o.ToTokenFile,
 		FromCAFile:    o.FromCAFile,
+		TLSCertFile:   o.TLSCertFile,
+		TLSKey:        o.TLSKey,
 
 		AnonymizeLabels:   o.AnonymizeLabels,
 		AnonymizeSalt:     o.AnonymizeSalt,
--- a/cmd/telemeter-client/main.go
+++ b/cmd/telemeter-client/main.go
@@ -51,7 +51,7 @@ func main() {
 
 	// TLS flags
 	cmd.Flags().StringVar(&opt.TLSCertFile, "tls-cert-file", "", "File containing the x509 public certificate to use when connecting to the Prometheus server.")
-	cmd.Flags().StringVar(&opt.TLSKey, "tls-private-key-file", "", "File containing the default x509 private key matching --tls-cert-file.")
+	cmd.Flags().StringVar(&opt.TLSKey, "tls-private-key-file", "", "File containing the  x509 private key to use when connecting to the Prometheus server. It should match with --tls-cert-file.")
 
 	cmd.Flags().StringVar(&opt.FromToken, "from-token", opt.FromToken, "A bearer token to use when authenticating to the source Prometheus server.")
 	cmd.Flags().StringVar(&opt.FromCAFile, "from-ca-file", opt.FromCAFile, "A file containing the CA certificate to use to verify the --from URL in addition to the system roots certificates.")
--- a/cmd/telemeter-client/main.go
+++ b/cmd/telemeter-client/main.go
@@ -50,7 +50,7 @@ func main() {
 	cmd.Flags().StringVar(&opt.From, "from", opt.From, "The Prometheus server to federate from.")
 
 	// TLS flags
-	cmd.Flags().StringVar(&opt.TLSCertFile, "tls-cert-file", "", "File containing the default x509 Certificate for HTTPS. (CA cert, if any, concatenated after server cert)")
+	cmd.Flags().StringVar(&opt.TLSCertFile, "tls-cert-file", "", "File containing the x509 public certificate to use when connecting to the Prometheus server.")
 	cmd.Flags().StringVar(&opt.TLSKey, "tls-private-key-file", "", "File containing the default x509 private key matching --tls-cert-file.")
 
 	cmd.Flags().StringVar(&opt.FromToken, "from-token", opt.FromToken, "A bearer token to use when authenticating to the source Prometheus server.")
--- a/cmd/telemeter-client/main.go
+++ b/cmd/telemeter-client/main.go
@@ -51,7 +51,7 @@ func main() {
 
 	// TLS flags
 	cmd.Flags().StringVar(&opt.TLSCertFile, "tls-cert-file", "", "File containing the x509 public certificate to use when connecting to the Prometheus server.")
-	cmd.Flags().StringVar(&opt.TLSKey, "tls-private-key-file", "", "File containing the  x509 private key to use when connecting to the Prometheus server. It should match with --tls-cert-file.")
+	cmd.Flags().StringVar(&opt.TLSKey, "tls-private-key-file", "", "File containing the x509 private key to use when connecting to the Prometheus server. It should match with --tls-cert-file.")
 
 	cmd.Flags().StringVar(&opt.FromToken, "from-token", opt.FromToken, "A bearer token to use when authenticating to the source Prometheus server.")
 	cmd.Flags().StringVar(&opt.FromCAFile, "from-ca-file", opt.FromCAFile, "A file containing the CA certificate to use to verify the --from URL in addition to the system roots certificates.")
--- a/jsonnet/telemeter/client/kubernetes.libsonnet
+++ b/jsonnet/telemeter/client/kubernetes.libsonnet
@@ -2,9 +2,12 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
 local secretName = 'telemeter-client';
 local secretVolumeName = 'secret-telemeter-client';
 local secretMountPath = '/etc/telemeter';
-local tlsSecret = 'metrics-client-certs';
+local tlsSecret = 'telemeter-client-certs';
 local tlsVolumeName = 'telemeter-client-tls';
 local tlsMountPath = '/etc/tls/private';
+local federateSecret = 'federate-client-certs';
+local federateVolumeName = 'federate-client-tls';
+local federateMountPath = '/etc/tls/private';
 local servingCertsCABundle = 'serving-certs-ca-bundle';
 local servingCertsCABundleFileName = 'service-ca.crt';
 local servingCertsCABundleMountPath = '/etc/%s' % servingCertsCABundle;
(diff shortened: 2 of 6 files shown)


pkg/forwarder/forwarder.go
```diff
@@ -148,7 +148,12 @@ func New(cfg Config) (*Worker, error) {
 
 	// Create the `fromClient`.
 	fromTransport := metricsclient.DefaultTransport()
-	if len(cfg.FromCAFile) > 0 {
+	fromClient := &http.Client{Transport: fromTransport}
+	if cfg.Debug {
+		level.Info(logger).Log("msg", "fromClient transport: Debug ")
```
danielmellado, line 153: If the transport is Debug, shouldn't Log level match that as well? I'd just move that.

pkg/forwarder/forwarder.go
```diff
@@ -148,7 +148,12 @@ func New(cfg Config) (*Worker, error) {
...
+		fromClient.Transport = telemeterhttp.NewDebugRoundTripper(logger, fromClient.Transport)
+	} else if len(cfg.FromCAFile) > 0 {
```
danielmellado, line 155: Not really important but maybe consider using a switch over here.

danielmellado (commented):
Besides my comments, would you mind adding some docs/release notes for this? Also consider whether this would need some test. Thanks!

pkg/forwarder/forwarder.go
marioferh: Right

marioferh:
> Besides my comments, would you mind adding some docs/release notes for this? Also consider whether this would need some test. Thanks!

About test, CI is testing it, waiting for the logs to check it.

About local test I am trying to refactor sh and add some local test before CI https://github.com/marioferh/telemeter/commit/dd5994c030bc02de70386643f6d6abdc6ef1676c

pkg/forwarder/forwarder.go
marioferh: done

jan--f: Not sure I follow this. Why do we not consider the TLS artifact configured of `Debug` is set?

simonpasquier:
hmm I think that we miss the options to pass the TLS certificate and key that should be used by the telemeter client when accessing the /federate endpoint.

pkg/forwarder/forwarder.go
```diff
@@ -146,12 +148,31 @@ func New(cfg Config) (*Worker, error) {
 		transformer.With(metricfamily.NewMetricsAnonymizer(anonymizeSalt, cfg.AnonymizeLabels, nil))
 	}
 
+	fmt.Println("forwawrder cafile from?=", cfg.FromCAFile)
```
danielmellado, line 151: check the typo over there ;)

pkg/forwarder/forwarder.go
```diff
@@ -148,10 +150,20 @@ func New(cfg Config) (*Worker, error) {
 
 	// Create the `fromClient`.
 	fromTransport := metricsclient.DefaultTransport()
+	fromClient := &http.Client{Transport: fromTransport}
+
 	if len(cfg.FromCAFile) > 0 {
-		if fromTransport.TLSClientConfig == nil {
-			fromTransport.TLSClientConfig = &tls.Config{}
+		level.Info(logger).Log("msg", "fromClient transport: CA file %s and TLS", cfg.FromCAFile)
```
simonpasquier, line 156:
```suggestion
		level.Info(logger).Log("msg", "TLS configuration", "ca_file", cfg.FromCAFile, "cert_file", cfg.TLSCertFile, "key_file", cfg.TLSKey)
```

pkg/forwarder/forwarder.go
```diff
@@ -148,10 +150,20 @@ func New(cfg Config) (*Worker, error) {
...
+
+		var cert tls.Certificate
+		var err error
+		if *&cfg.TLSCertFile != "" && *&cfg.TLSKey != "" {
```
simonpasquier, line 160:
```suggestion
		if cfg.TLSCertFile != "" && cfg.TLSKey != "" {
```

pkg/forwarder/forwarder.go
```diff
@@ -148,10 +150,20 @@ func New(cfg Config) (*Worker, error) {
...
 	fromTransport := metricsclient.DefaultTransport()
+	fromClient := &http.Client{Transport: fromTransport}
+
 	if len(cfg.FromCAFile) > 0 {
-		if fromTransport.TLSClientConfig == nil {
-			fromTransport.TLSClientConfig = &tls.Config{}
+		level.Info(logger).Log("msg", "fromClient transport: CA file %s and TLS", cfg.FromCAFile)
+
+		var cert tls.Certificate
+		var err error
```
simonpasquier, line 159: to be safe in case metricsclient.DefaultTransport() happens to initialize the TLSClientConfig in the future.
```suggestion
		var err error

		if fromTransport.TLSClientConfig == nil {
			fromTransport.TLSClientConfig = &tls.Config{}
		}
```

pkg/forwarder/forwarder.go
```diff
@@ -163,22 +175,35 @@ func New(cfg Config) (*Worker, error) {
...
-			return nil, fmt.Errorf("unable to read from-token-file: %v", err)
+
+		fromTransport.TLSClientConfig = &tls.Config{
+			Certificates: []tls.Certificate{cert},
+			RootCAs:      pool,
+		}
+
+		fromClient.Transport = telemeterhttp.NewTLSRoundTripper(fromClient.Transport)
+
+	} else {
```
simonpasquier, line 186: I would prefer to support multiple auth configurations at the same time rather being exclusive. E.g. if client cert/key files and token bearer are configured at the same time then setup the TLS configuration + the bearer round tripper. Same reasoning for the debug round-tripper, it could be useful even when using TLS cert/key only.

pkg/http/roundtripper.go
```diff
@@ -37,6 +37,18 @@ func NewDebugRoundTripper(logger log.Logger, next http.RoundTripper) *debugRound
...
+}
+
+func NewTLSRoundTripper(next http.RoundTripper) *TLSRoundTripper {
+	return &TLSRoundTripper{next}
+}
+
+func (rt *TLSRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	return rt.wrapper.RoundTrip(req)
+}
+
```
simonpasquier, line 51: Is this really needed?

cmd/telemeter-client/main.go
```diff
@@ -48,6 +48,11 @@ func main() {
 
 	cmd.Flags().StringVar(&opt.Listen, "listen", opt.Listen, "A host:port to listen on for health and metrics.")
 	cmd.Flags().StringVar(&opt.From, "from", opt.From, "The Prometheus server to federate from.")
+
+	// TLS flags
+	cmd.Flags().StringVar(&opt.TLSCertFile, "tls-cert-file", "", "File containing the default x509 Certificate for HTTPS. (CA cert, if any, concatenated after server cert)")
```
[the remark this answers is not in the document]
marioferh: Sorry c&p

pkg/forwarder/forwarder.go
marioferh: Resolve in next comments

pkg/http/roundtripper.go
marioferh: There is NewBearerRoundTripper and DebugRoundTripper, it is needed a new function to create a different one, right?

pkg/forwarder/forwarder.go
marioferh: done

pkg/http/roundtripper.go
marioferh: done

raptorsun:
/retitle MON-3230: Add TLS auth to telemeter-client

cmd/telemeter-client/main.go
```diff
@@ -48,6 +48,11 @@ func main() {
 
 	cmd.Flags().StringVar(&opt.Listen, "listen", opt.Listen, "A host:port to listen on for health and metrics.")
 	cmd.Flags().StringVar(&opt.From, "from", opt.From, "The Prometheus server to federate from.")
+
+	// TLS flags
+	cmd.Flags().StringVar(&opt.TLSCertFile, "tls-cert-file", "", "File containing the x509 public certificate to use when connecting to the Prometheus server.")
+	cmd.Flags().StringVar(&opt.TLSKey, "tls-private-key-file", "", "File containing the  x509 private key to use when connecting to the Prometheus server. It should match with --tls-cert-file.")
```
raptorsun, line 54:
```suggestion
	cmd.Flags().StringVar(&opt.TLSKey, "tls-private-key-file", "", "File containing the x509 private key to use when connecting to the Prometheus server. It should match with --tls-cert-file.")
```

pkg/forwarder/forwarder.go
```diff
@@ -163,20 +179,26 @@ func New(cfg Config) (*Worker, error) {
...
+			RootCAs:      pool,
+		}
 	}
-	fromClient := &http.Client{Transport: fromTransport}
 	if cfg.Debug {
+		level.Info(logger).Log("msg", "fromClient transport: Debug ")
 		fromClient.Transport = telemeterhttp.NewDebugRoundTripper(logger, fromClient.Transport)
 	}
 	if len(cfg.FromToken) == 0 && len(cfg.FromTokenFile) > 0 {
+		level.Info(logger).Log("msg", "fromClient transport FromTokenFile")
```
raptorsun, line 193: Is this log for debugging purposes indicating entry of this code block or for showing the file name of `cfg.FromTokenFile`?

raptorsun: In that case we can use `level.Debug()` . 
When debugging we pass `--log-level=debug`, when in normal use it does not show up.

raptorsun: same for line 201.
other looks good to me :)

marioferh: done, thanks!

pkg/forwarder/forwarder.go
```diff
@@ -163,20 +179,26 @@ func New(cfg Config) (*Worker, error) {
 		if !pool.AppendCertsFromPEM(data) {
 			level.Warn(logger).Log("msg", "no certs found in from-ca-file")
 		}
-		fromTransport.TLSClientConfig.RootCAs = pool
+
+		fromTransport.TLSClientConfig = &tls.Config{
+			Certificates: []tls.Certificate{cert},
+			RootCAs:      pool,
+		}
```
simonpasquier, line 186: It would be better to set only the fields that we specifically want to update rather than replacing the full struct.

```suggestion

		fromTransport.TLSClientConfig.Certificates = []tls.Certificate{cert}
		fromTransport.TLSClientConfig.RootCAs = pool
```

pkg/forwarder/forwarder.go
```diff
@@ -163,20 +179,26 @@ func New(cfg Config) (*Worker, error) {
...
 	}
-	fromClient := &http.Client{Transport: fromTransport}
 	if cfg.Debug {
+		level.Debug(logger).Log("msg", "fromClient transport: Debug ")
```
simonpasquier, line 189:
```suggestion
		level.Debug(logger).Log("msg", "enabling the debug round tripper for the fromClient transport")
```

pkg/forwarder/forwarder.go
```diff
@@ -163,20 +179,26 @@ func New(cfg Config) (*Worker, error) {
...
 		fromClient.Transport = telemeterhttp.NewDebugRoundTripper(logger, fromClient.Transport)
 	}
 	if len(cfg.FromToken) == 0 && len(cfg.FromTokenFile) > 0 {
+		level.Debug(logger).Log("msg", "fromClient transport FromTokenFile")
```
simonpasquier, line 193:
```suggestion
		level.Debug(logger).Log("msg", "enabling the token file round tripper for the fromClient transport")
```

pkg/forwarder/forwarder.go
```diff
@@ -163,20 +179,26 @@ func New(cfg Config) (*Worker, error) {
...
 		data, err := ioutil.ReadFile(cfg.FromTokenFile)
 		if err != nil {
 			return nil, fmt.Errorf("unable to read from-token-file: %v", err)
 		}
 		cfg.FromToken = strings.TrimSpace(string(data))
 	}
 	if len(cfg.FromToken) > 0 {
+		level.Debug(logger).Log("msg", "fromClient transport FromToken")
```
simonpasquier, line 201:
```suggestion
		level.Debug(logger).Log("msg", "enabling the token round tripper for the fromClient transport")
```

simonpasquier (commented):
Just a few nits.

marioferh:
> Just a few nits.

thanks, much better like that

jsonnet/telemeter/client/kubernetes.libsonnet
```diff
@@ -119,6 +124,8 @@ local securePort = 8443;
           '/usr/bin/telemeter-client',
           '--id=$(ID)',
           '--from=$(FROM)',
+          '--tls-cert-file=/etc/tls/private/tls.crt',
+          '--tls-private-key-file=/etc/tls/private/tls.key',
           '--from-ca-file=%s/%s' % [servingCertsCABundleMountPath, servingCertsCABundleFileName],
           '--from-token-file=' + fromTokenFile,
```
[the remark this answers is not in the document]
marioferh: Yep, it is working and tested in local, but if I remove from-token-file arg I am removing token file test. I need to find the way to test both scenarios right?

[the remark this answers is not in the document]
simonpasquier: My initial idea was that we would configure the telemeter client with the TLS certificate/key only. The downside is that if the certificate becomes invalid (e.g. CMO doesn't rotate it) then telemetry will stop working which isn't a nice experience.
As for testing I've got no good idea how we can ensure that TLS is effectively used vs. the Prometheus kube-rbac-proxy falling back to token authentication...

PR #461, opened 2023-05-10
Fix integration test documentation

marioferh:
Test documentation is old and not accurate. 

Command ./test/integration.sh has path of binaries broken, better use "make test-integration"
No args are passed to integration.sh


test/integration.sh
```diff
@@ -2,9 +2,6 @@
 
 # Runs a semi-realistic integration test with two servers, a stub authorization server, a 
 # prometheus that scrapes from them, and a single client that fetches "cluster" metrics.
```
marioferh, line 4: if required I can add the args to work with other ports and tokens but I think is not neccesary

PR #464, opened 2023-06-02
OCPBUGS-12345: tools: update to golangci-lint v1.51

jan--f:
Seems like versions <=1.50.0 can cause memory issues with golang 1.20.

This also pulls in #460.


jan--f:
/retitle OCPBUGS-12345: tools: update to golangci-lint v1.51

PR #469, opened 2023-06-23
WIP: Remove token to test tls configuration

marioferh:
Remove token to connect with TLS between telemeter and prometheus.


marioferh:
close, just for testing

PR #471, opened 2023-06-27
Revert "MON-3230: Add TLS auth to telemeter-client"

marioferh:
Reverts openshift/telemeter#455

Secret name should be the same


danielmellado:
Can't the secret name be changed without having to revert the whole PR?

marioferh:
> Can't the secret name be changed without having to revert the whole PR?

yep

PR #476, opened 2023-08-17
RHOBS-855: rhelemeter: Read client info/data from request

philipgough:
https://issues.redhat.com/browse/RHOBS-855

### DIFF
--- a/cmd/rhelemeter-server/main_e2e_test.go
+++ b/cmd/rhelemeter-server/main_e2e_test.go
@@ -7,7 +7,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net"
 	"net/http"
 	"net/http/httptest"
@@ -56,7 +56,7 @@ var expectedTimeSeries = []prompb.TimeSeries{
 	},
 }
 
-func TestServerRHEL(t *testing.T) {
+func TestServerRhelMtls(t *testing.T) {
 	defer goleak.VerifyNone(t)
 
 	receiveServer := httptest.NewServer(mockedReceiver(t))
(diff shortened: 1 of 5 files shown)


pkg/authorize/ssl/ssl.go
```diff
@@ -0,0 +1,60 @@
...
+
+// ClientCertInfoAsHeaders is middleware that extracts client information from the request and validates
+// the pre-shared key
+func ClientCertInfoAsHeaders(config ClientCertConfig, logger log.Logger) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		fn := func(w http.ResponseWriter, r *http.Request) {
+			secret := r.Header.Get(config.Config.SecretHeader)
+			if secret == "" {
+				http.Error(w, fmt.Sprintf("secret must be sent in request header %s",
+					config.Config.SecretHeader), http.StatusForbidden)
```
thibaultmg, line 48: Small detail, I wonder wether 401 would be more appropriate as here no credentials are provided?

thibaultmg (approved):
Awesome!
Reading at the related conversation I see that we do not yet check if the system is still registered by calling Candlepin. I guess this is expected?

pkg/authorize/ssl/ssl.go
philipgough: yeah that probably makes more sense. In the absence of a cert the server seems to return a 403 so I followed that but I think its fine to return 401

cmd/rhelemeter-server/main_e2e_test.go
```diff
@@ -159,6 +159,166 @@ func TestServerRHEL(t *testing.T) {
 	}
 }
 
+func TestServerRhelWithClientInfoFromHeaders(t *testing.T) {
+	defer goleak.VerifyNone(t)
+
+	receiveServer := httptest.NewServer(mockedReceiver(t))
+	defer receiveServer.Close()
```
douglascamata, line 166: Small suggestion here: might want to use `t.Cleanup(receiveServer.Close)`, as a `defer` won't run panic, but the first will.

douglascamata (approved):
Just a very small and non-blocking suggestion, LGTM! 🚀

philipgough:
Thanks to you both for reviews. I need to add jsonnet changes assuming this all works in the end so will take care of both comments in said follow up.

moadz (approved):
LGTM

PR #477, opened 2023-09-08
RHOBS-870: Improve logging

thibaultmg:
Improve logging for incident resolution by:

- Adding a request logger
- Adding request id in all logs (was not consistently done)
- Adding some missing logs

To be noted: As a result of some simplifications, following error messages are now returned to the client from `pkg/authorize/jwt/client_authorizer.go`:
```
func (j *clientAuthorizer) AuthorizeClient(tokenData string) (*authorize.Client, error) {
	if !j.hasCorrectIssuer(tokenData) {
		return nil, false, nil
	}

	tok, err := jwt.ParseSigned(tokenData)
	if err != nil {
		return nil, false, nil
	}
```

I don't see it as an issue. In fact, quite the opposite. Let me know if you disagree.

### DIFF
--- a/pkg/authorize/handler.go
+++ b/pkg/authorize/handler.go
@@ -12,29 +12,33 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/go-chi/chi/middleware"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 )
 
-func NewAuthorizeClientHandler(authorizer ClientAuthorizer, next http.Handler) http.Handler {
+func NewAuthorizeClientHandler(logger log.Logger, authorizer ClientAuthorizer, next http.Handler) http.Handler {
+	logger = log.With(logger, "component", "authorize")
+
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		logger := log.With(logger, "request", middleware.GetReqID(req.Context()))
+
 		auth := strings.SplitN(req.Header.Get("Authorization"), " ", 2)
 		if strings.ToLower(auth[0]) != "bearer" {
 			http.Error(w, "Only bearer authorization allowed", http.StatusUnauthorized)
+			level.Debug(logger).Log("msg", "Only bearer authorization allowed", "auth", auth[0])
 			return
 		}
 		if len(auth) != 2 || len(strings.TrimSpace(auth[1])) == 0 {
 			http.Error(w, "Invalid Authorization header", http.StatusUnauthorized)
+			level.Debug(logger).Log("msg", "Invalid Authorization header", "auth", auth)
 			return
 		}
 
-		client, ok, err := authorizer.AuthorizeClient(auth[1])
+		client, err := authorizer.AuthorizeClient(auth[1])
 		if err != nil {
 			http.Error(w, fmt.Sprintf("Not authorized: %v", err), http.StatusUnauthorized)
-			return
-		}
-		if !ok {
-			http.Error(w, "Not authorized", http.StatusUnauthorized)
+			level.Warn(logger).Log("msg", "Not authorized", "err", err)
 			return
 		}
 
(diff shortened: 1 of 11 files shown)


pkg/authorize/handler.go
```diff
@@ -146,12 +152,18 @@ func NewHandler(logger log.Logger, client *http.Client, endpoint *url.URL, tenan
...
+		logger = log.With(logger, "tenant", tenant)
+
 		if _, err := AgainstEndpoint(logger, client, endpoint, token, tenant, nil); err != nil {
 			level.Warn(logger).Log("msg", "unauthorized request made:", "err", err)
 			w.WriteHeader(http.StatusUnauthorized)
 			return
 		}
 
+		// Log the tenant for debugging purposes.
+		level.Info(logger).Log("msg", "authorized request")
```
philipgough, line 165: should this be debug I wonder? might be spammy at info?

pkg/authorize/tollbooth/tollbooth.go
```diff
@@ -67,12 +65,10 @@ func (a *authorizer) AuthorizeCluster(token, cluster string) (string, error) {
 
 	response := &clusterRegistration{}
 	if err := json.Unmarshal(body, response); err != nil {
-		level.Warn(a.logger).Log("msg", "upstream server response could not be parsed", "to", a.to)
 		return "", fmt.Errorf("unable to parse response body: %v", err)
 	}
 
 	if len(response.AccountID) == 0 {
-		level.Warn(a.logger).Log("msg", "upstream server responded with an empty user string", "to", a.to)
```
philipgough, line 75: do we log these now in the caller?

philipgough (commented):
looks good overall with some questions and comments

pkg/authorize/handler.go
thibaultmg: The idea here is to have the `tenant` being logged alongside the request log. Thinking about the incident we had a few weeks ago, I though it would be great to have this data available at Info level (as well as the clusterID). And I cannot get it easily at the request logger level. So I have put it here. But yes, might be spammy.
I'll put debug level as suggested.

pkg/authorize/jwt/handler.go
```diff
@@ -50,48 +54,64 @@ func (a *authorizeClusterHandler) ServeHTTP(w http.ResponseWriter, req *http.Req
...
 		uid := rand.Int63()
-		level.Error(a.logger).Log("msg", "unable to authorize request", "uid", uid, "err", err)
+		level.Warn(logger).Log("msg", "unable to authorize request", "err", err, "uid", uid)
 		http.Error(w, fmt.Sprintf("Internal server error, requestid=%d", uid), http.StatusInternalServerError)
 		return
 	}
 
+	// Log the subject for debugging purposes
+	logger = log.With(logger, "subject", subject)
+	level.Info(logger).Log("msg", "authorized request")
```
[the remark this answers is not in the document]
thibaultmg: Same reply 😉

pkg/authorize/tollbooth/tollbooth.go
thibaultmg: Yes, any error returned gets logged. I wanted to simplify logging a bit. Logs are done at every level here. It is difficult to reason about. When adding a log somewhere, you don't know if it has already been logged in called functions. Now, only the caller of this function do logs.

pkg/cache/cache.go
```diff
@@ -36,41 +37,50 @@ type RoundTripper struct {
...
+			level.Error(logger).Log("msg", "failed to read response from cache", "err", err)
+		}
+
+		return resp, errors.Wrap(err, "failed to read response from cache")
 	}
 
 	r.cacheReadsTotal.WithLabelValues("miss").Inc()
 	resp, err := r.next.RoundTrip(req)
-	if err == nil && resp.StatusCode/200 == 1 {
+	if err == nil && resp.StatusCode >= 200 && resp.StatusCode < 300 {
```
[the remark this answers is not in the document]
thibaultmg: Previous check did catch 3xx range. I don't think it was intentional isn't?

philipgough (approved):

philipgough (approved):

PR #479, opened 2023-09-14
Fix: handle authorization server 403 responses

thibaultmg:
As a result of a previous change, 403 returned by the authorization server now result in 500 returned to the client. This triggers retries and alerts from our monitoring system.
This PR reverts to the previous behaviour.


saswatamcode (approved):
LGTM!

philipgough (approved):
/lgtm

PR #491, opened 2023-10-17
rhelemeter: Validation of incoming metrics

philipgough:
The rhelemeter server exposes remote write API and this change allows us to do some further validation on the incoming requests. 

Right now, the flow of traffic is a many-to-one relationship between hosts/VMs running RHEL OS and rhelemeter-server.
In some cases, an intermediate proxy will validate and terminate mTLS and in those cases, the proxy must forward on the client information, as well as a pre-shared key in a set of headers (configurable) like so

```
x-gateway-secret: ${PSK}  //Example PSK -> 46545757575
x-certauth-cn:  ${Certificate Subject}  //Example  Certificate Subject ->  "/O = 12345, /CN = 1234567-9b7c-48f6-8d5b-7654321
```

This change enables us to decode the Subjects CommonName and Organisation in middleware and set it on the context of the request to be dealt with further in the chain of handlers. This is done in such a fashion since we already decode the remote write request as part of the label and metric whitelisting logic. This avoids us having to bare the cost of decoding and wrapping the request again in such fashion with the view of keeping the hot path as optimised as possible, since we piggy-back on the existing parsed time series.

The drawback to this is the slight added complexity but my view is that with an ever growing number of incoming requests expected, this is the preferred option.

### DIFF
--- a/cmd/rhelemeter-server/README.md
+++ b/cmd/rhelemeter-server/README.md
@@ -1,0 +1,75 @@
+# Rhelemeter
+
+Rhelemeter is a Prometheus remote write forward proxy. It allows RHEL hosts or an intermediate proxy to push metrics to
+a upstream Prometheus (or Prometheus remote write compatible) endpoint.
+
+[rhelemeter-server](https://github.com/openshift/telemeter/tree/master/cmd/rhelemeter-server) needs to receive
+and send metrics across security boundaries, and thus performs additional authorization and data integrity checks as required.
+
+Architecture
+```mermaid
+C4Deployment
+    title rhelemeter-server
+
+    Deployment_Node(rhel-os, "RHEL OS VM", ""){
+        Container(rhel, "Prometheus Remote Write Client", "", "Writes host OS based metrics.")
+    }
+
+    Deployment_Node(proxy, "Intermediate Proxy", ""){
+        Container(remote-write-proxy, "mTLS auth proxy", "", "Validates client certificate and forwards contents as header")
+    }
+        
+    Deployment_Node(rhelemeter, "rhelemeter-server", "Running in Red Hat infrastructure"){
+        Container(rhelemeter-server, "rhelemeter-server", "", "Validates sender and request contents and forwards upstream or rejects.")
+    }
+        
+    Deployment_Node(observatorium, "observatorium", ""){
+        Container(observatorium, "observatorium-api", "", "Performs authz and ingests metrics")
+    }
+        
+    Deployment_Node(sso, "SSO", "") {
+        Container(rh-sso, "Red Hat SSO", "", "Issues and validates tokens")
+    }
+        
+
+    Rel(rhel, remote-write-proxy, "Remote writes to", "")
+    Rel(remote-write-proxy, rhelemeter-server, "Forwards requests to", "")
+    Rel(rhelemeter-server, observatorium, "Forwards requests to", "")
+    Rel_R(rhelemeter-server, rh-sso, "Fetches token from")
+    Rel_R(observatorium, rh-sso, "Validates token with")
+
+    UpdateRelStyle(rhel, remote-write-proxy, $offsetX="-40", $offsetY="-30")   
+    UpdateRelStyle(remote-write-proxy, rhelemeter-server, $offsetY="-40")
+    UpdateRelStyle(rhelemeter-server, observatorium, $offsetY="-20", $offsetX="-50")
+    UpdateRelStyle(api, rh-sso, $offsetX="-40", $offsetY="-20")
+    UpdateRelStyle(observatorium, rh-sso, $offsetY="+40")
+```
+
+## Overview
+
+The server on startup fetches a token from the configured authorization server and uses that to forward
+requests to the upstream remote write endpoint. Tokens are refreshed automatically when they expire.
+
+The server supports two modes of operation:
+
+### Intermediate auth proxy
+This is the flow described in the diagram above. The proxy is responsible for validating the client certificate and
+forwarding the contents of the request as a headers. The server then validates the header using a pre-shared key
+and forwards the request to the upstream as configured.
+
+The `client-info-data-file` flag is used to specify the header names used to extract client info data.
+
+
+Optionally, the `client-info-subject-label` flag can be used to specify a label name to store the client subject.
+If present, the labels value on each time series much match the common name as forwarded by the proxy.
+If set, the server expects the header to be present in the following format where `x-rh-certauth-cn` 
+is configurable as described above:
+
+```bash
+x-rh-certauth-cn:  ${Certificate Subject}  
+// Example  Certificate Subject ->  "/O = 1234567, /CN = 1234567-9b7c-48f6-8d5b-7654321```
+```
+
+### mTLS
+The server is responsible for validating and decoding the client certificate and determining the client identity.
+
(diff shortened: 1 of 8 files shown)


pkg/authorize/ssl/ssl.go
```diff
@@ -1,11 +1,19 @@
...
 	"net/http"
+	"strings"
 
 	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+)
+
+const (
+	OrganizationContextKey = "org"
+	CommonNameContextKey   = "common_name"
```
douglascamata, line 16: Not a good practice to use strings as context keys. We can use something like `struct{}` (empty struct) to avoid allocations, among other options.

douglascamata (changes requested):
Awesome stuff, just one small request. 🙇

pkg/authorize/ssl/ssl.go
douglascamata: Reference: https://pkg.go.dev/context#WithValue

pkg/authorize/ssl/ssl.go
```diff
@@ -1,13 +1,19 @@
...
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 )
 
+type OrganizationContextKey struct{}
+type CommonNameContextKey = struct{}
```
douglascamata, line 15:
```suggestion
type CommonNameContextKey struct{}
```

cmd/rhelemeter-server/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+Rhelemeter is a Prometheus remote write forward proxy. It allows RHEL hosts or an intermediate proxy to push metrics to
+a upstream Prometheus (or Prometheus remote write compatible) endpoint.
+
+[rhelemeter-server](https://github.com/openshift/telemeter/tree/master/cmd/rhelemeter-server) needs to receive
+and send metrics across security boundaries, and thus performs additional authorization and data integrity checks as required.
+
+Architecture
+```mermaid
+C4Deployment
```
thibaultmg, line 11: That's so awesome ❤️

cmd/rhelemeter-server/README.md
```diff
@@ -0,0 +1,75 @@
...
+    title rhelemeter-server
+
+    Deployment_Node(rhel-os, "RHEL OS VM", ""){
+        Container(rhel, "Prometheus Remote Write Client", "", "Writes host OS based metrics.")
+    }
+
+    Deployment_Node(proxy, "Intermediate Proxy", ""){
```
thibaultmg, line 18: I would name the proxy, would be easier to understand. `Intermediate Proxy (Akamai)`

cmd/rhelemeter-server/README.md
```diff
@@ -0,0 +1,75 @@
...
+        Container(remote-write-proxy, "mTLS auth proxy", "", "Validates client certificate and forwards contents as header")
+    }
+        
+    Deployment_Node(rhelemeter, "rhelemeter-server", "Running in Red Hat infrastructure"){
+        Container(rhelemeter-server, "rhelemeter-server", "", "Validates sender and request contents and forwards upstream or rejects.")
+    }
+        
+    Deployment_Node(observatorium, "observatorium", ""){
```
thibaultmg, line 26: We could also name the instance here `observatorium-mst`

pkg/receive/handler.go
```diff
@@ -294,6 +307,24 @@ func (h *Handler) TransformAndValidateWriteRequest(next http.Handler, labels ...
 			return
 		}
 
+		for _, ts := range wreq.GetTimeseries() {
```
thibaultmg, line 310: We could avoid entering these nested loops when no label value needs to be enforced.

pkg/authorize/ssl/ssl.go
```diff
@@ -52,7 +63,57 @@ func ClientCertInfoAsHeaders(config ClientCertConfig, logger log.Logger) func(ht
...
+					config.Config.CommonNameHeader), http.StatusForbidden)
+				return
+			}
+
+			// cnInfo must be provided in the format of "/O = xyz, /CN = 123
+			parts := strings.Split(cnInfo, ",")
+			if len(parts) != 2 {
+				level.Info(logger).Log("msg", "invalid format for organisation and CN", "cnInfo", cnInfo)
+				http.Error(w, fmt.Sprintf("invalid format for organisation and CN %s",
+					config.Config.CommonNameHeader), http.StatusForbidden)
```
thibaultmg, line 78: Is there a mistake in the error string? I guess you wanted `invalid format for organisation and CN in http header %s`

pkg/authorize/ssl/ssl.go
```diff
@@ -52,7 +63,57 @@ func ClientCertInfoAsHeaders(config ClientCertConfig, logger log.Logger) func(ht
...
+				return
+			}
+
+			ctx := r.Context()
+			for i, part := range parts {
+				subPart := strings.Split(part, "=")
```
thibaultmg, line 84: Why aren't you using a regex for all these checks? Something like `/O = (?P<Org>[^,]+), /CN = (?P<CN>[^,]+)`

pkg/authorize/ssl/ssl_test.go
```diff
@@ -53,10 +79,29 @@ func TestClientCertInfoAsHeaders(t *testing.T) {
 			},
 			expect: http.StatusForbidden,
 		},
+		{
+			name: "Test missing CN returns 403",
```
thibaultmg, line 83: We're not testing the missing CN here. We are being rejected because the secret is not set. I think that for these tests, we should hace a base request that works from which we are removing parts to make it selectively fail.

thibaultmg:
Left some small comments. Thanks for the documentation with the graph, it's awesome! Makes the repo more approachable.

cmd/rhelemeter-server/README.md
philipgough: I purposely left this lacking lots of detail. The intention was more to describe config and traffic flow. We have a requirement to created a detailed internal architecture document for this and other components so I would prefer to push it to there. Does that make sense?

cmd/rhelemeter-server/README.md
philipgough: As above. This describes some "observatorium" backend but I'm not convinced we need to be more detailed in this repo.

pkg/receive/handler.go
philipgough: Good catch. Thanks

pkg/authorize/ssl/ssl.go
philipgough: for this I'd be inclined to just push out what we have and tidy up in a follow up if that is ok? I still haven't been able to see an actual request as routed through the proxy so the format is still unclear to me as of now

cmd/rhelemeter-server/README.md
thibaultmg: Yes makes sense 👍

thibaultmg (approved):

pkg/receive/handler.go
```diff
@@ -73,6 +75,12 @@ func NewHandler(logger log.Logger, forwardURL string, client *http.Client, reg p
 				Help: "The number of remote write requests dropped due to missing labels.",
 			},
 		),
+		requestIncorrectLabelValues: prometheus.NewCounter(
+			prometheus.CounterOpts{
+				Name: "telemeter_receive_request_incorrect_label_values_total",
```
moadz, line 80:
```suggestion
				Name: "telemeter_receive_requests_incorrect_label_values_total",
```

moadz (approved):
LGTM

PR #492, opened 2023-10-23
rhelemeter: Parse org and cn with regex

philipgough:


pkg/authorize/ssl/ssl.go
```diff
@@ -49,6 +50,8 @@ func (conf ClientCertConfig) Validate() error {
 // ClientCertInfoAsHeaders is middleware validates the pre-shared key and
 // extracts client information from the request and adds it to the request context
 func ClientCertInfoAsHeaders(config ClientCertConfig, logger log.Logger) func(http.Handler) http.Handler {
+	var commonNameRE = regexp.MustCompile(`/CN=([a-zA-Z0-9-]*$)`)
+	var organisationRE = regexp.MustCompile(`/O=([a-zA-Z0-9-]*)`)
```
thibaultmg, line 54: I would even include the next `/` to make sure we grab the full Org or nothing is badly formatted: `/O=([a-zA-Z0-9-]*)/`

thibaultmg (approved):
Small non blocking comment. Looks good!

PR #504, opened 2024-02-07
OCPBUGS-16397: update nutanix filter

racedo:
to catch nutanix_ahv which is what the cluster:virt_platform_nodes:sum expression assumes


simonpasquier (commented):
/lgtm

PR #506, opened 2024-02-08
RHOBS-995: Simplify cluster:capacity_effective_cpu_cores, add tests

kahowell:
Simplify by dividing all x86_64 cpu counts by 2.

Note that this takes advantage of the way that the SKUs are structured, where the capacity is written as multiples of "2 cores or 4vCPUs".

One difference in how this simplification works is that with nodes reporting more than 2 threads-per-core will be counted by CPUs, rather than by cores.

When exactly 2 threads-per-core are reported, there is no functional difference, as node_role_os_version_machine:cpu_capacity_cores:sum already divides CPUs by 2.

This adds testing similar to what's in cluster-monitoring-operator, covering only the `cluster:capacity_effective_cpu_cores` rule.

I had to update the prometheus version, as promtool was too old and incorrectly flagging existing rules.

I added a note about rule tests to the README.

I did not update the prow config, because I don't know where to, but happy to do an update for that given some hints.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ UP_BIN=$(BIN_DIR)/up
 OTELCOL_BIN=$(BIN_DIR)/otelcol
 MEMCACHED_BIN=$(BIN_DIR)/memcached
 PROMETHEUS_BIN=$(BIN_DIR)/prometheus
+PROMTOOL_BIN=$(BIN_DIR)/promtool
 GOJSONTOYAML_BIN=$(BIN_DIR)/gojsontoyaml
 JSONNET_BIN?=$(BIN_DIR)/jsonnet
 JSONNETFMT_BIN?=$(BIN_DIR)/jsonnetfmt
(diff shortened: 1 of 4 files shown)


kahowell:
d851af5a..f7039dce: rebase

Makefile
```diff
@@ -20,9 +20,10 @@ UP_BIN=$(BIN_DIR)/up
 OTELCOL_BIN=$(BIN_DIR)/otelcol
 MEMCACHED_BIN=$(BIN_DIR)/memcached
 PROMETHEUS_BIN=$(BIN_DIR)/prometheus
+PROMTOOL_BIN=$(BIN_DIR)/promtool
 GOJSONTOYAML_BIN=$(BIN_DIR)/gojsontoyaml
 JSONNET_BIN?=$(BIN_DIR)/jsonnet
-JSONNETFMT_BIN?=$(BIN_DIR)/jsonnetfmt
+SONNETFMT_BIN?=$(BIN_DIR)/jsonnetfmt
```
simonpasquier, line 26:
```suggestion
JSONNETFMT_BIN?=$(BIN_DIR)/jsonnetfmt
```

simonpasquier:
> I did not update the prow config, because I don't know where to, but happy to do an update for that given some hints.

I'll follow-up once we get this merged :)

jsonnet/telemeter/rules.libsonnet
```diff
@@ -171,23 +171,18 @@
               // OpenShift Cluster effective cores for subscription usage.
               // This counts both worker nodes and, when the control plane is schedulable, control plane nodes.
               // Only CoreOS nodes are counted.
-              // 1. x86_64 nodes that show hyperthreading in the telemetry have an accurate cores value in node_role_os_version_machine:cpu_capacity_cores:sum.
-              // 2. x86_64 nodes that do not show hyperthreading need the cores value adjusted to account for 2 threads per core (* 0.5).
-              // 3. Other CPU architectures are assumed to have accurate values in node_role_os_version_machine:cpu_capacity_cores:sum.
+              // 1. x86_64 nodes need the cores value adjusted to account for 2 threads per core (* 0.5).
+              // 2. Other CPU architectures are assumed to have accurate values in cluster:capacity_cpu_cores:sum.
               record: 'cluster:capacity_effective_cpu_cores',
```
simonpasquier, line 176: since the recording expression only uses `cluster:capacity_cpu_cores:sum`, shouldn't we also modify  `cluster:cpu_capacity_cores:_id` to use the same metric? It would also simplify the test inputs since the tests would properly generate `cluster:capacity_cpu_cores:_id`.

Makefile
kahowell: Oops. Accidental edit, will amend this into the current commit to limit noise in commit history.

simonpasquier (commented):
/lgtm
/hold
@kahowell looks good to me, thanks!
I've put a `/hold` on the PR so you can merge it when you feel it's ready from your side.

simonpasquier:
I'll investigate separately why the integration job fails (it's unrelated to this PR).

simonpasquier:
> > I did not update the prow config, because I don't know where to, but happy to do an update for that given some hints.
> 
> I'll follow-up once we get this merged :)

https://github.com/openshift/release/pull/48786

PR #525, opened 2024-03-14
MON-3737: Update binaries to new version

marioferh:
Change pinned memcached binary downloaded in makefile. Related to this PR https://github.com/openshift/release/pull/50465 that install memcached as a dnf dependency.
memcached binary was download from archlinux archive and pinned to an old image that create some issues in integration test.

Update otelcol to last version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,6 @@ GOLANGCI_LINT_VERSION=v1.56.2
 THANOS_BIN=$(BIN_DIR)/thanos
 UP_BIN=$(BIN_DIR)/up
 OTELCOL_BIN=$(BIN_DIR)/otelcol
-MEMCACHED_BIN=$(BIN_DIR)/memcached
 PROMETHEUS_BIN=$(BIN_DIR)/prometheus
 PROMTOOL_BIN=$(BIN_DIR)/promtool
 GOJSONTOYAML_BIN=$(BIN_DIR)/gojsontoyaml
@@ -168,7 +167,7 @@ test-generate:
 test-format:
 	make --always-make format && git diff --exit-code
 
-test-integration: build | $(THANOS_BIN) $(UP_BIN) $(MEMCACHED_BIN) $(PROMETHEUS_BIN)
+test-integration: build | $(THANOS_BIN) $(UP_BIN) $(PROMETHEUS_BIN)
 	@echo "================================="
 	@echo ">>> Running integration tests: V1"
 	@echo "================================="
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -223,18 +223,18 @@ $(LIB_DIR):
 
 $(MEMCACHED_BIN): | $(BIN_DIR) $(LIB_DIR)
 	@echo "Downloading Memcached"
-	curl -L https://archive.org/download/archlinux_pkg_libevent/libevent-2.1.10-1-x86_64.pkg.tar.xz | tar --strip-components=2 --xz -xf - -C $(LIB_DIR) usr/lib
-	curl -L https://archive.org/download/archlinux_pkg_memcached/memcached-1.5.8-1-x86_64.pkg.tar.xz | tar --strip-components=2 --xz -xf - -C $(BIN_DIR) usr/bin/memcached
+	curl -L https://archive.org/download/archlinux_pkg_libevent/libevent-2.1.12-1-x86_64.pkg.tar.zst | tar --strip-components=2 --use-compress-program=unzstd -xf - -C $(LIB_DIR) usr/lib
```
simonpasquier, line 226: ideally we should install from RPM if available

marioferh: where could I get the rpm's?

simonpasquier: In the CI env, they might be available from Red Hat repositories. You could installing them from dnf if the container image is based on rhel?

danielmellado: You can modify the Dockerfile in a way like as this one
```
RUN dnf install -y --setopt=tsflags=nodocs memcached memcached-selinux && \
    dnf clean all && rm -rf /var/cache/*
```

marioferh: I can try again with dnf but my last attempts it was failing due to libraries compatibily

marioferh: But, it is needed Memcached in local to run with the scripts. Even if we installed memcached in the container, we would need another test container to run the test. This is feasible but it will take time, upload the images, change the ci, it is worth?

Makefile
```diff
@@ -220,10 +220,9 @@ $(BIN_DIR):
 $(LIB_DIR):
 	mkdir -p $@
 
-$(MEMCACHED_BIN): | $(BIN_DIR) $(LIB_DIR)
-	@echo "Downloading Memcached"
-	curl -L https://archive.org/download/archlinux_pkg_libevent/libevent-2.1.12-1-x86_64.pkg.tar.zst | tar --strip-components=2 --use-compress-program=unzstd -xf - -C $(LIB_DIR) usr/lib
-	curl -L https://archive.org/download/archlinux_pkg_memcached/memcached-1.6.18-1-x86_64.pkg.tar.zst | tar --strip-components=2 --use-compress-program=unzstd -xf - -C $(BIN_DIR) usr/bin/memcached
+$(MEMCACHED_BIN): 
+	@echo "test dnf" 
+	dnf install memcached-1.6.21-2.fc39.x86_64
```
danielmellado, line 225: don't install in that way. I'd leave the specific version out as that may slightly change in the future and at least add user `dnf install -y --setopt=tsflags=nodocs`

Otherwise it'll require confirmation and most likely fail and also pull unused docs and waste space.

marioferh: done

philipgough (approved):
/lgtm

PR #529, opened 2024-04-12
RHOBS-1036:Harden integration tests

philipgough:
This change merges the two integration tests files. The only real difference between these files was the memcached dependency which is included now as default. 

This removes the dependency on `up`, preferring the writes to come from telemeter client which is a more valuable integration test IMHO.   

Dependencies are brought up in order and tested for readiness before moving on.
We set the default tenant on Thanos receive to avoid having to initialise a TSDB at request time (would result in a series of 5xx).


marioferh (commented):
/lgtm

Thanks, I think now is much better and clearer, and just one script it will be easier to maintain.

PR #518, opened 2024-03-12
Create rule for rosa HCP vCPU-hours

kahowell:
For ROSA HCP we need to consider only metrics coming from the management cluster (ignoring metrics coming from the data plane)

PR #544, opened 2024-09-24
OSD-25491,OSD-25492:New recording rules averaging CPU consumption for ROSA HCP & ROSA classic clusters

Nikokolas3270:
Those recording rules will be used for the (new) billing of ROSA

Closes [OSD-25491](https://issues.redhat.com//browse/OSD-25491) and [OSD-25492](https://issues.redhat.com//browse/OSD-25492)

### DIFF
--- a/jsonnet/telemeter/rules.libsonnet
+++ b/jsonnet/telemeter/rules.libsonnet
@@ -204,6 +204,15 @@
                 max by (_id, managed_cluster_id) (acm_managed_cluster_worker_cores:max)
               |||,
             },
+            {
+              // Hypershift Cluster vCPU-hours for the last hour.
+              // The divisor of scalar(count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
+              // sum(...) by (_id) is used ensure a single datapoint per cluster ID.
+              record: 'hostedcluster:hypershift_cluster_vcpus:vcpu_hours',
+              expr: |||
+                max by(_id)(sum_over_time(hostedcluster:hypershift_cluster_vcpus:max[1h:5m])) / scalar(count_over_time(vector(1)[1h:5m]))
+              |||,
+            },
           ],
         },
       ],
--- a/jsonnet/telemeter/rules.libsonnet
+++ b/jsonnet/telemeter/rules.libsonnet
@@ -213,6 +213,13 @@
                 max by(_id)(sum_over_time(hostedcluster:hypershift_cluster_vcpus:max[1h:5m])) / scalar(count_over_time(vector(1)[1h:5m]))
               |||,
             },
+            {
+              // ROSA (HCP or classic) Cluster vCPU-hours for the last hour.
+              record: 'rosa:cluster:vcpu_hours',
+              expr: |||
+                hostedcluster:hypershift_cluster_vcpus:vcpu_hours or on (_id) cluster:usage:workload:capacity_virtual_cpu_hours
+              |||,
+            },
           ],
         },
       ],
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ telemeter-bench
 /bin
 /benchmark
 benchmark.pdf
+/tmp
 
 # These are empty target files, created on every docker build. Their sole
 # purpose is to track the last target execution time to evalualte, whether the
--- a/test/rulestests.yaml
+++ b/test/rulestests.yaml
@@ -157,3 +157,54 @@ tests:
                   value: 32
                 - labels: 'acm_capacity_effective_cpu_cores{_id="another_hub_cluster",managed_cluster_id="none_ocp_aks"}'
                   value: 32
+    # hostedcluster:hypershift_cluster_vcpus:vcpu_hours tests
+    - interval: 1m
+      input_series:
+          - series: 'hostedcluster:hypershift_cluster_vcpus:max{_id="my-id", _mc_id="my-mc-id", mac_name="my-mc-name", exported_namespace="ocm-some-env-my-id"}'
+            values: '0x90 4x90 0x90'
+      promql_expr_test:
+          - expr: hostedcluster:hypershift_cluster_vcpus:vcpu_hours
+            eval_time: 0
+            exp_samples:
+                - labels: 'hostedcluster:hypershift_cluster_vcpus:vcpu_hours{_id="my-id"}'
+                  value: 0
+          - expr: hostedcluster:hypershift_cluster_vcpus:vcpu_hours
+            eval_time: 60m
+            exp_samples:
+                - labels: 'hostedcluster:hypershift_cluster_vcpus:vcpu_hours{_id="my-id"}'
+                  value: 0
+          - expr: hostedcluster:hypershift_cluster_vcpus:vcpu_hours
+            eval_time: 90m
+            exp_samples:
+                - labels: 'hostedcluster:hypershift_cluster_vcpus:vcpu_hours{_id="my-id"}'
+                  value: 0
+          - expr: hostedcluster:hypershift_cluster_vcpus:vcpu_hours
+            eval_time: 122m
+            exp_samples:
+                - labels: 'hostedcluster:hypershift_cluster_vcpus:vcpu_hours{_id="my-id"}'
+                  value: 2
+          - expr: hostedcluster:hypershift_cluster_vcpus:vcpu_hours
+            eval_time: 152m
+            exp_samples:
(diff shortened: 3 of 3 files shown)


simonpasquier (commented):
Can we have unit tests for the new rules like we had in #518 and #506?

saswatamcode (approved):
LGTM

Nikokolas3270:
Unit-tests added as requested by Simon

jsonnet/telemeter/rules.libsonnet
```diff
@@ -204,6 +204,22 @@
                 max by (_id, managed_cluster_id) (acm_managed_cluster_worker_cores:max)
               |||,
             },
+            {
+              // Hypershift Cluster vCPU-hours for the last hour.
+              // The divisor of scalar(count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
+              // sum(...) by (_id) is used ensure a single datapoint per cluster ID.
+              record: 'hostedcluster:hypershift_cluster_vcpus:vcpu_hours',
```
simonpasquier, line 211: (suggestion) "...vcpus:vcpu..." feels redundant. Also https://prometheus.io/docs/practices/rules/#naming
```suggestion
              record: 'hostedcluster:hypershift_cluster_vcpus:avg1h',
```

jsonnet/telemeter/rules.libsonnet
```diff
@@ -204,6 +204,22 @@
                 max by (_id, managed_cluster_id) (acm_managed_cluster_worker_cores:max)
               |||,
             },
+            {
+              // Hypershift Cluster vCPU-hours for the last hour.
```
simonpasquier, line 208: HyperShift == ROSA HCP? Since we mention "ROSA HCP" below, it might be good to use the same terminology here.

jsonnet/telemeter/rules.libsonnet
```diff
@@ -204,6 +204,22 @@
...
+              // The divisor of scalar(count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
+              // sum(...) by (_id) is used ensure a single datapoint per cluster ID.
+              record: 'hostedcluster:hypershift_cluster_vcpus:vcpu_hours',
+              expr: |||
+                max by(_id)(sum_over_time(hostedcluster:hypershift_cluster_vcpus:max[1h:5m])) / scalar(count_over_time(vector(1)[1h:5m]))
+              |||,
+            },
+            {
+              // ROSA (HCP or classic) Cluster vCPU-hours for the last hour.
```
simonpasquier, line 217: Maybe worth noting that we prefer the ROSA HCP metric if both exist for the same cluster _id?

simonpasquier (commented):
/lgtm
/hold
I have some cosmetic remarks but nothing blocking (especially since I'm not using these metrics directly).

simonpasquier:
/retitle OSD-25491,OSD-25492:New recording rules averaging CPU consumption for ROSA HCP & ROSA classic clusters

simonpasquier:
@Nikokolas3270 feel free to unblock if you want (again my comments were nice-to-have).

jsonnet/telemeter/rules.libsonnet
```diff
@@ -204,6 +204,22 @@
                 max by (_id, managed_cluster_id) (acm_managed_cluster_worker_cores:max)
               |||,
             },
+            {
+              // Hypershift Cluster vCPU-hours for the last hour.
+              // The divisor of scalar(count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
+              // sum(...) by (_id) is used ensure a single datapoint per cluster ID.
+              record: 'hostedcluster:hypershift_cluster_vcpus:vcpu_hours',
+              expr: |||
+                max by(_id)(sum_over_time(hostedcluster:hypershift_cluster_vcpus:max[1h:5m])) / scalar(count_over_time(vector(1)[1h:5m]))
```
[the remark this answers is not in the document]
Nikokolas3270: `scalar(count_over_time(vector(1)[1h:5m])` is not always equals to 12. It oscillate between 12 and 13.
Also, even it was always equal to 12, replacing it by `12` will feel like a hard code; not to mention that several other recording rules already use the same kind of construct in this file.

jsonnet/telemeter/rules.libsonnet
Nikokolas3270: Naming `hostedcluster:hypershift_cluster_vcpus:vcpu_hours` recording rule that way has been requested in [OSD-25491](https://issues.redhat.com/browse/OSD-25491). I agree that the name is not ideal (repeating the `vcpu` part...); that said it is consistent with the other recording rules in the file "averaging" CPU consumption during the last hour.
Also, naming this recording rule that way is quite useful to make sure that `rosa:cluster:vcpu_hours` recording rule, based on it, is not comparing apples and oranges.
I am not against changing this suffix, but if we do it or start to do it, lets make sure it is consistently changed everywhere.

jsonnet/telemeter/rules.libsonnet
Nikokolas3270: Well that's a tricky question.
Telemeter "hosts" metrics for all kind of clusters, including unmanaged clusters.
For this reason, the comment on top of `cluster:usage:workload:capacity_virtual_cpu_hours` definition is not saying that it concerns ROSA classic only... Because in fact this recording rule is defined and takes values for all kind of clusters, including ROSA classic but also OSD clusters or even OCP clusters installed on customer premises.
The "sister" `hostedcluster:hypershift_cluster_vcpus:vcpu_hours` recording rule suffers from the same fate: it is defined on cluster running the hypershift operator. This includes "our" management clusters hosting the ROSA HCP clusters but this could include as well any customer cluster running the hypershift operator.
So Hypershift != ROSA HCP.
With that in mind, this raises the concern that recording rule `rosa:cluster:vcpu_hours` is maybe poorly named. Indeed this rule is defined for:
- OSD clusters which are not ROSA
- OCP clusters installed and managed by the customers
- Clusters hosted by Hypershift but which are managed by the customer.

jsonnet/telemeter/rules.libsonnet
Nikokolas3270: Well, I think that the expression is pretty clear about that.
But I agree that this may bother people not familiar with Prometheus expression. I will add a comment in that direction.

simonpasquier: From experience, the more information we add here the more future ourselves will be grateful :)

jsonnet/telemeter/rules.libsonnet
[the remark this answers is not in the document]
simonpasquier: > `scalar(count_over_time(vector(1)[1h:5m])` is not always equals to 12. It oscillate between 12 and 13.

ok now I remember this glitch in Thanos. In theory, it should always return 12 which is why I suggested to hardcode but indeed it's not possible.

simonpasquier (commented):
/hold cancel

My remaining remarks are only about code comments, no need to block the PR on them.

jsonnet/telemeter/rules.libsonnet
Nikokolas3270: Yeah, I have started a thread with the author of the JIA ticket.
We should eventually consider refactoring those rules from time to time and better name the rules ans the metrics.

jsonnet/telemeter/rules.libsonnet
```diff
@@ -204,6 +204,22 @@
                 max by (_id, managed_cluster_id) (acm_managed_cluster_worker_cores:max)
               |||,
             },
+            {
+              // Hypershift Cluster vCPU-hours for the last hour.
+              // The divisor of scalar(count_over_time(vector(1)[1h:5m])) allows us to get an effective average value having "absent data" treated as 0.
+              // sum(...) by (_id) is used ensure a single datapoint per cluster ID.
```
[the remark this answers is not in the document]
Nikokolas3270: Corrective PR:
https://github.com/openshift/telemeter/pull/546

jsonnet/telemeter/rules.libsonnet
Nikokolas3270: Corrective PR:
https://github.com/openshift/telemeter/pull/546

jsonnet/telemeter/rules.libsonnet
kahowell: FWIW, swatch is a single consumer of this metric, and we can easily handle whatever name is most clear to others. So I'm inclined to name precisely based on what the metric is, rather than its intended use.

PR #550, opened 2024-11-19
MON-4059: split TelemeterClientFailures into send/retrieve alerts

slashpai:
Replace `TelemeterClientFailures` with two alerts on `metricsclient_http_requests_total`:
- **TelemeterClientSendErrors** (`federate_to`) — upload failures to telemeter
- **TelemeterClientRetrieveErrors** (`federate_from`) — federation failures from Prometheus
Both fire when non-2xx rate > 20% for 1h. Descriptions distinguish 4xx (cluster-admin actionable) from 5xx (Red Hat service).

Ref: https://github.com/openshift/runbooks/blob/master/alerts/cluster-monitoring-operator/TelemeterClientFailures.md#diagnosis

### DIFF
--- a/jsonnet/telemeter/client/kubernetes.libsonnet
+++ b/jsonnet/telemeter/client/kubernetes.libsonnet
@@ -252,23 +252,52 @@ local securePort = 8443;
                 expr: 'max(federate_samples - federate_filtered_samples)',
               },
               {
+                alert: 'TelemeterClientSendErrors',
                 expr: |||
-                  sum by (namespace) (
-                    rate(federate_requests_failed_total{job="telemeter-client"}[15m])
-                  ) /
-                  sum by (namespace) (
-                    rate(federate_requests_total{job="telemeter-client"}[15m])
+                  (
+                    sum by (namespace) (rate(metricsclient_http_requests_total{job="telemeter-client",client="federate_to",status_code!~"2.."}[15m]))
+                    /
+                    sum by (namespace) (rate(metricsclient_http_requests_total{job="telemeter-client",client="federate_to"}[15m]))
                   ) > 0.2
                 |||,
+                'for': '1h',
                 labels: {
                   severity: 'warning',
                 },
                 annotations: {
-                  description: 'The telemeter client in namespace {{ $labels.namespace }} fails {{ $value | humanize }} of the requests to the telemeter service.\nCheck the logs of the telemeter-client pod with the following command:\noc logs -n openshift-monitoring deployment.apps/telemeter-client -c telemeter-client\nIf the telemeter client fails to authenticate with the telemeter service, make sure that the global pull secret is up to date, see https://docs.openshift.com/container-platform/latest/openshift_images/managing_images/using-image-pull-secrets.html#images-update-global-pull-secret_using-image-pull-secrets for more details.',
-                  summary: 'Telemeter client fails to send metrics',
+                  summary: 'Telemeter client is failing to send metrics.',
+                  description: |||
+                    The telemeter client in namespace {{ $labels.namespace }} has {{ $value | humanizePercentage }} error rate when sending metrics to the telemeter service.
+                    4xx errors typically indicate an authentication or authorization issue — check that the global pull secret is valid and up to date:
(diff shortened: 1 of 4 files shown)


jsonnet/telemeter/client/kubernetes.libsonnet
```diff
@@ -253,12 +253,7 @@ local securePort = 8443;
               },
               {
                 expr: |||
-                  sum by (namespace) (
-                    rate(federate_requests_failed_total{job="telemeter-client"}[15m])
-                  ) /
-                  sum by (namespace) (
-                    rate(federate_requests_total{job="telemeter-client"}[15m])
-                  ) > 0.2
+                  sum by(client, status_code) (rate(metricsclient_http_requests_total{status_code!~"200"}[15m])) > 0
```
simonpasquier, line 256: (suggestion) we need to preserve the `namespace` label. Let's also evaluate the rate of failed requests.
```suggestion
                  sum by(client, status_code,namespace) (rate(metricsclient_http_requests_total{status_code!~"2..",job="telemeter-client"}[15m]))
                  / 
                  on(client, namespace) group_left() sum by(client, namespace) (rate(metricsclient_http_requests_total{job="telemeter-client"}[15m])) > 0.2
```

simonpasquier (commented):
I'd recommend to create 2 alerting rules: one for the `federate_from` client and another one for the `federate_to` client. They can have the same name but different descriptions.

slashpai:
Tested the alert `TelemeterClientSendErrors` by patching telemeter client deployment in clusterbot cluster by replacing  `--to=$(TO)` with separate endpoints 

```
--to-auth
--to-upload
```

 with wrong value set for `--to-upload` and reducing alert time to test firing

Screenshot 2026-09-09 at 12 03 51 PM


To test `TelemeterClientRetrieveErrors`, deleted role binding `telemeter-client-view`, reduced alert time to test firing and restarted telemeter deployment

Screenshot 2026-09-09 at 12 40 19 PM

slashpai:
@simonpasquier I have revived this PR and updated per suggestions. Included tests and locally tested as above.

Take a look when you get chance

jsonnet/telemeter/client/kubernetes.libsonnet
```diff
@@ -252,23 +252,52 @@ local securePort = 8443;
...
-                  summary: 'Telemeter client fails to send metrics',
+                  summary: 'Telemeter client is failing to send metrics.',
+                  description: |||
+                    The telemeter client in namespace {{ $labels.namespace }} has {{ $value | humanizePercentage }} error rate when sending metrics to the telemeter service.
+                    4xx errors typically indicate an authentication or authorization issue — check that the global pull secret is valid and up to date:
+                      oc get secret pull-secret -n openshift-config -o jsonpath='{.data.\.dockerconfigjson}' | base64 -d
+                    5xx errors indicate a server-side issue on Red Hat telemeter service and are not actionable by the cluster administrator.
```
simonpasquier, line 273: 🤔 do we even need to alert in this case? maybe we just need an info alert for 5xx responses.

simonpasquier (commented):
what do we do with the runbook? obviously we need to keep it for supported OCP releases but we can either extend it to cover the new alerting rules or create dedicated runbooks for each (my preference would be option 2 as it's more obvious for customers and we could eventually remove the old runbook).

jsonnet/telemeter/client/kubernetes.libsonnet
slashpai: info is good.

slashpai:
> what do we do with the runbook? obviously we need to keep it for supported OCP releases but we can either extend it to cover the new alerting rules or create dedicated runbooks for each (my preference would be option 2 as it's more obvious for customers and we could eventually remove the old runbook).

Ya. Since we can 

> what do we do with the runbook? obviously we need to keep it for supported OCP releases but we can either extend it to cover the new alerting rules or create dedicated runbooks for each (my preference would be option 2 as it's more obvious for customers and we could eventually remove the old runbook).

We can add new ones since we currently one runbook per alert

PR #558, opened 2025-05-27
OCPBUGS-56425: Accept and discard metadata requests over remote write

philipgough:


pkg/receive/handler.go
```diff
@@ -229,6 +237,18 @@ func (h *Handler) TransformAndValidateWriteRequest(next http.Handler, labels ...
 			return
 		}
 
+		if len(wreq.GetTimeseries()) == 0 && len(wreq.GetMetadata()) > 0 {
+			// thanos does not yet support metadata requests, so we silently accept and discard them here.
+			// we may want to add a flag to enable this in the future when thanos can consume metadata requests via remote write.
+			h.metadataRequestsTotal.Inc()
+
+			w.WriteHeader(http.StatusOK)
```
moadz, line 245: Is this how we manage the status on the write request? What if it fails after telemeter for any other reason - would this misreport a 200?

moadz (approved):
/lgtm

pkg/receive/handler.go
philipgough: this is a case of exiting early if we dont support the request. we want to accept it and discard it in the case of metadata only. protected by the if statement above it.

pkg/receive/handler.go
```diff
@@ -229,6 +237,18 @@ func (h *Handler) TransformAndValidateWriteRequest(next http.Handler, labels ...
...
+			_ = json.NewEncoder(w).Encode(map[string]string{
+				"message": "metadata requests are not supported yet, silently discarding",
```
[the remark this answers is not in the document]
philipgough: if they choose to log the body then yes, helps to debug whats happening client side. key point to note is that you will really only fall in here if you are sending metadata so would typically be from a prom client (problem surfaced by hypershift for example). 

telemeter client wont be sending metadata so they wont fall in here.

PR #564, opened 2025-07-14
OCPBUGS-16397: Add id_primary_host_product rule to Nutanix telemetry

danielmellado:
Add new id_primary_host product recording rule for reliable platform
detection. It classifies now Nutanix clusters as host_products='nutanix'
instead of random kvm/virt-unknown.

OCPBUGS-16397: Nutanix clusters inconsistently classified in telemetry

Signed-off-by: Daniel Mellado <dmellado@fedoraproject.org>

### DIFF
--- a/docs/nutanix-telemtry-fix.md
+++ b/docs/nutanix-telemtry-fix.md
@@ -1,0 +1,151 @@
+# Nutanix Telemetry Classification Fix
+
+## Overview
+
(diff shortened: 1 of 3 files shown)


racedo:
Please note that if the provider is None and the host type is virt-unknown or kvm_unknown we don't know if the cluster is indeed a nutanix cluster, so any workaround will miss these clusters

PR #566, opened 2025-10-20
OCPBUGS-62765: port ACM recording rules to OCP 4.16

elgnay:
Port ACM recording rule `acm_capacity_effective_cpu_cores` to OCP 4.16, as it is currently available only in OCP 4.17 and later.
The related issue: https://issues.redhat.com/browse/OCPBUGS-62765


simonpasquier:
/hold

Server-side recording rules are deployed from main/master. There's no need to update them in release branches.

PR #578, opened 2026-02-17
SREP-3493: add `cluster:usage:control_plane:instance_hours`

typeid:
This PR reverts changes to `cluster:usage:workload:capacity_physical_instance_hours` from https://github.com/openshift/telemeter/pull/577 and moves our changes to a new recording rule `cluster:usage:control_plane:instance_hours`.

### DIFF
--- a/jsonnet/telemeter/rules.libsonnet
+++ b/jsonnet/telemeter/rules.libsonnet
@@ -158,22 +158,14 @@
               |||,
             },
             {
-              // OpenShift Control Plane instance hours over the past hour.
-              // This recording rule represents the instance hours over the past hour in which the control plane has been active.
-              // This recording rule is based on two metrics:
-              // 1) `cluster:usage:workload:capacity_physical_cpu_cores:max:5m` which is used as an anchor for the control plane hours in non-HCP clusters.
-              // 2) `hostedcluster:hypershift_cluster_vcpus:max` which is used as an anchor for the control plane hours in HCP clusters.
-              // While these metrics relate to the actual worker count, they are emitted by either infrastructure nodes for non-HCP clusters, 
-              // or the management cluster layer for HCP clusters and at least 1 time series is expected at any point in time during the cluster's lifetime.
-              // 
-              // Note for HCP: 
-              // While for HCP both metrics may be emitted to telemetry (one from the telemeter-client, one from the management cluster), 
-              // this is not always the case, as the telemeter-client is optional or may not be reliable because it runs on the worker nodes.
-              // We fall back to the more reliable metric emitted from the management cluster in any case. 
+              // OpenShift Cluster Instance-hours for the last hour.
               record: 'cluster:usage:workload:capacity_physical_instance_hours',
               expr: |||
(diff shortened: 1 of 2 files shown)


jsonnet/telemeter/rules.libsonnet
```diff
@@ -170,13 +170,20 @@
               // While for HCP both metrics may be emitted to telemetry (one from the telemeter-client, one from the management cluster),
               // this is not always the case, as the telemeter-client is optional or may not be reliable because it runs on the worker nodes.
               // We fall back to the more reliable metric emitted from the management cluster in any case.
-              record: 'cluster:usage:workload:capacity_physical_instance_hours',
+              record: 'cluster:usage:control_plane:instance_hours',
               expr: |||
                 max by(_id) (count_over_time(cluster:usage:workload:capacity_physical_cpu_cores:max:5m[1h:5m])) / scalar(steps:count1h)
```
simonpasquier: is this the same as `cluster:usage:workload:capacity_physical_instance_hours` below? if yes, can we move this recording rule after and leverage `cluster:usage:workload:capacity_physical_instance_hours`?

simonpasquier (commented):
Can we add unit tests at some point to avoid regressions?

typeid:
/hold

Let me incorporate Simon's comments :)

typeid:
@simonpasquier I've updated the PR, but I'm keeping the hold for now. Discussing with BU if they are okay dropping `cluster:usage:workload:capacity_physical_instance_hours` from the new control plane expression as it is unnecessary and might bite us in the future. 

The reason we kept it for now is to not interfere with existing billing, but we are confident fully relying on `max by(_id) (count_over_time((hostedcluster:hypershift_cluster_vcpus:max > 0)[1h:5m])) / scalar(steps:count1h)` is good enough.

barnabycourt:
It seems risky to change the definition of cluster:usage:workload:capacity_physical_instance_hours in place. Shouldn't we be validating the new rule against the old rule all the way to production before changing the definition of the old rule?

typeid:
@barnabycourt as mentioned in slack already, we're not changing `cluster:usage:workload:capacity_physical_instance_hours` - it is the same as it currently is on telemeter prod after this change.

@simonpasquier  / @saswatamcode we should be good to merge now.

simonpasquier:
/lgtm

I assume that control plane telemetry is more reliable than telemetry from the data plane. Do we envision switching the priority between both at some point?

Also should we monitor the difference between the metrics when both exist?

typeid:
> I assume that control plane telemetry is more reliable than telemetry from the data plane. Do we envision switching the priority between both at some point?

That is correct. For now the business decision was to default to the current billing and fall back to the control plane billing to have the least changes in billing.

In an ideal state we would only use the control plane telemetry. 

I'll defer to @jiri-fiala on the question of when/if we envision changing the priority / removing the data plane metric at some point.

> Also should we monitor the difference between the metrics when both exist?

I believe this task is already assigned. We will test the difference between the two recording rules ahead of swapping over billing. :)

simonpasquier:
/label acknowledge-critical-fixes-only

the change doesn't land in the OCP core payload.

PR #579, opened 2026-02-19
SREP-3493: deduplicate cluster:usage:control_plane:instance_hours output series

typeid:
This PR fixes the `cluster:usage:control_plane:instance_hours` recording rule to not attempt and emit duplicate time series.

The `or` can result in two series found by `_id`, which would result in the recording rule creating duplicate time series, which is not allowed.

It's unclear to me why the unit tests didn't catch this, but on staging telemeter we get the following issue: `vector contains metrics with the same labelset after applying rule labels`.

### DIFF
--- a/jsonnet/telemeter/rules.libsonnet
+++ b/jsonnet/telemeter/rules.libsonnet
@@ -179,11 +179,9 @@
               // We fall back to the more reliable metric emitted from the management cluster in any case.
               record: 'cluster:usage:control_plane:instance_hours',
               expr: |||
-                max by (_id) (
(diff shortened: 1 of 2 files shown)


saswatamcode (approved):
/lgtm

simonpasquier:
/retitle [SREP-3493](https://issues.redhat.com/browse/SREP-3493): deduplicate cluster:usage:control_plane:instance_hours output series

simonpasquier:
/label  acknowledge-critical-fixes-only 

Telemeter server change, no change to the OCP payload.

PR #581, opened 2026-03-11
SREP-3493: update cluster:usage:workload:capacity_physical_instance_hours

typeid:
This PR updates `cluster:usage:workload:capacity_physical_instance_hours` to match the previous`cluster:usage:control_plane:instance_hours`.

SubWatch [is currently consuming cluster:usage:workload:capacity_physical_instance_hours](https://github.com/RedHatInsights/rhsm-subscriptions/blob/56473057412b56f818d258d445a519203e0dc320/swatch-product-configuration/src/main/resources/subscription_configs/OpenShift/rosa.yaml#L36), and by updating the expression they don't have to go through a release process.

The idea is that this goes through staging gated by telemeter staging, then to production once there's green light from the team. At a later point in time, SubWatch will swap directly to `cluster:usage:control_plane:instance_hours`.

## Summary by CodeRabbit

* **Refactor**
  * Improved instance-hour calculations by adding an alternate vCPU-based measurement for hosted clusters and consolidating control-plane instance-hour reporting to include this alternative alongside the existing workload-capacity source.
* **Chores**
  * Added explanatory TODOs and reference notes marking the alternate computation for future removal and cleanup.

### DIFF
--- a/jsonnet/telemeter/rules.libsonnet
+++ b/jsonnet/telemeter/rules.libsonnet
@@ -159,9 +159,17 @@
             },
             {
               // OpenShift Cluster Instance-hours for the last hour.
+              // TODO(typeid): We will remove this rule at a later point as part of SREP-3493
+              // and replace it with `cluster:usage:control_plane:instance_hours`.
+              // This rule is currently equivalent to `cluster:usage:control_plane:instance_hours` to allow
+              // bumping this through stage and production without a SubWatch release.
+              // SubWatch currently references this rule for ROSA HCP control plane billing via
+              // https://github.com/RedHatInsights/rhsm-subscriptions/blob/1309ecd73f21cc213f5ca58b5486ff95fe8ee982/swatch-product-configuration/src/main/resources/subscription_configs/OpenShift/rosa.yaml#L46
               record: 'cluster:usage:workload:capacity_physical_instance_hours',
               expr: |||
                 max by(_id) (count_over_time(cluster:usage:workload:capacity_physical_cpu_cores:max:5m[1h:5m])) / scalar(steps:count1h)
+                or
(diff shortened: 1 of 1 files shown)


simonpasquier (commented):
Just to confirm: `cluster:usage:workload:capacity_physical_instance_hours` and `cluster:usage:control_plane:instance_hours` have now identical expressions and will return the same results?
If yes can we make it the second recording rule a simple alias to the first?

````
- record: first
  expr: ...
- record: second
  expr: first
````

Can we also add a comment with a link to the SREP ticket and a TODO?

typeid:
/hold 

Waiting on BU's +1 to go ahead.

typeid:
@simonpasquier Thanks for the review. I added a todo with an explanation.

> Just to confirm: cluster:usage:workload:capacity_physical_instance_hours and cluster:usage:control_plane:instance_hours have now identical expressions and will return the same results?
If yes can we make it the second recording rule a simple alias to the first?

The only reason I didn't do this is that I don't want to additionally modify `cluster:usage:control_plane:instance_hours` once we remove `cluster:usage:workload:capacity_physical_instance_hours`. I don't think there's more than a very minor risk to swapping it to a different expression later, but if it isn't a blocker I would prefer to keep those fully identical.

simonpasquier:
thanks for the TODO, I was mostly concerned about future ourselves wondering about why the 2 rules look like they are.

You've got a lint issue to fix.

typeid:
Apologies for the extra pushes, the `make format` has some issues with BSD grep (Perl-style negative lookahead (?!...) is not supported).

jiri-fiala:
adding +1 - this is what we asked for

simonpasquier (commented):
/lgtm

PR #582, opened 2026-03-26
OCPBUGS-79699: rules: add rhel to label_node_openshift_io_os_id

danielmellado:
4.19+ reports label_node_openshift_io_os_id=rhel while remaining CoreOS;
the previous rhcos-only selector dropped cluster:capacity_effective_cpu_cores
from telemetry. This commit adds rhel to the rule and a test case.

Signed-off-by: Daniel Mellado <dmellado@fedoraproject.org>

### DIFF
--- a/jsonnet/telemeter/rules.libsonnet
+++ b/jsonnet/telemeter/rules.libsonnet
@@ -202,25 +202,25 @@
               // returns 0 for any cluster reporting core capacity, used to improve performance of cluster:capacity_effective_cpu_cores
               record: 'cluster:cpu_capacity_cores:_id',
               expr: |||
-                group by(_id, tenant_id) (cluster:capacity_cpu_cores:sum{label_node_openshift_io_os_id="rhcos"}) * 0
+                group by(_id, tenant_id) (cluster:capacity_cpu_cores:sum{label_node_openshift_io_os_id=~"rhcos|rhel"}) * 0
               |||,
             },
             {
               // OpenShift Cluster effective cores for subscription usage.
               // This counts both worker nodes and, when the control plane is schedulable, control plane nodes.
-              // Only CoreOS nodes are counted.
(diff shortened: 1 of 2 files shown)


jsonnet/telemeter/rules.libsonnet
```diff
@@ -202,25 +202,25 @@
               // returns 0 for any cluster reporting core capacity, used to improve performance of cluster:capacity_effective_cpu_cores
               record: 'cluster:cpu_capacity_cores:_id',
               expr: |||
-                group by(_id, tenant_id) (cluster:capacity_cpu_cores:sum{label_node_openshift_io_os_id="rhcos"}) * 0
+                group by(_id, tenant_id) (cluster:capacity_cpu_cores:sum{label_node_openshift_io_os_id=~"rhcos|rhel"}) * 0
```
juzhao, line 205: `label_node_openshift_io_os_id=~"rhcos|rhel"}` is safe
I think since 4.19+ the os id is changed to rhel now, how about leave rhel only
`label_node_openshift_io_os_id="rhel"})`, same for others

danielmellado: Thanks for the review! If we stop backporting this in 4.19+ then that'd be fine. Using only label_node_openshift_io_os_id="rhel" would fix 4.19+ but would stop matching clusters still on 4.18 and earlier, where the label remains rhcos. 

My main concern here is that telemeter rules run against the full fleet, so we may  need both values. The regex keeps backward compatibility with the original rhcos filter and adds rhel for 4.19+, wdyt?

juzhao: I see, you are right, since the telemeter server side configuration only have main branch for production, example: https://github.com/rhobs/configuration/blob/main/resources/services/rhobs-thanos-operator/staging/telemeter-rules.yaml#L149-L160, keep rhcos is right for 4.18 and below, add rhel is for 4.19+, and seems we also need to update in the rhobs for the same after this PR merged

juzhao:
thanks, LGTM, waiting for others to review

PR #588, opened 2026-04-28
OCPBUGS-84383: fix(deps): bump go-jose/v3 to v3.0.5 for JWE unwrap DoS

marioferh:
Upgrade github.com/go-jose/go-jose/v3 from v3.0.3 to v3.0.5.

Earlier releases could panic when decrypting a crafted JWE whose alg is a key-wrapping algorithm (suffix KW, excluding GCM-KW variants) and encrypted_key is empty or too short. KeyUnwrap then tried to allocate a slice with invalid length, which could crash the process (denial of service). v3.0.5 rejects too-short ciphertext and returns an error instead.

Regenerated vendor with go mod vendor.

Made-with: Cursor

## Summary by CodeRabbit

* **Chores**
  * Updated dependency versions.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/coreos/go-oidc/v3 v3.9.0
 	github.com/efficientgo/core v1.0.0-rc.2
 	github.com/go-chi/chi v4.0.4+incompatible
-	github.com/go-jose/go-jose/v3 v3.0.3
+	github.com/go-jose/go-jose/v3 v3.0.5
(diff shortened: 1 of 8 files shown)


simonpasquier:
/lgtm
/verified by tests
/jira refresh

Note that the issue doesn't affect telemeter client which doesn't import `github.com/go-jose/go-jose/v3` (only the server binary does).

danielmellado:
Note that we should also backport this to the release-4.22 and 4.23 branches.

PR #589, opened 2026-04-30
[release-4.21] OCPBUGS-80868: Bump google.golang.org/grpc to v1.79.3

ocp-sustaining-admins:
This is an automated PR from the ARC system.

**Go version used:** `go1.24.13`

**Modules affected:** `.`, `tools`

The commands used to generate this PR were:

```
go get google.golang.org/grpc@v1.79.3
go mod tidy
go mod vendor
```

> **Note:** This package bump updated the `go` directive in `go.mod`. This may cause CI check failures if the repository's CI configuration does not yet support the updated Go version.

A member of the Red Hat Openshift Sustaining Team will review the PR and take appropriate action.


simonpasquier:
we need to fix the Jira chain (e.g. OCPBUGS-80868 doesn't depend on a ticket targeting 4.22.0): https://github.com/openshift/telemeter/pull/589#issuecomment-4377963095

germanparente:
Label set as per Simon former comment that there should not a dependency to check.

simonpasquier:
The lint job fails because of Go version mismatch

```
Error: can't load config: the Go language version (go1.23) used to build golangci-lint is lower than the targeted Go version (1.24.13)
```

The first option is to also bump the golangci-lint version (but it might trigger other issues to fix). The second option is to override the job (it is okay-ish since we don't modify the code directly).

Atharva-Shinde:
I'll leave the decision to you @simonpasquier. Is this a pressing issue or is it okay to override?

PR #605, opened 2026-05-07
OCPBUGS-84021: update google.golang.org/grpc

simonpasquier:


simonpasquier:
/override lint

We have incompatible Go version requirements between golangci-lint (needs 1.25) and the builder image (1.24).

PR #611, opened 2026-06-30
OCPBUGS-93760: bump github.com/prometheus/prometheus

simonpasquier:


moadz (approved):
LGTM, OOI why do we check in `/vendor`? Do we do offline builds? Or is this just an OCP standard for quicker/reliable builds?

simonpasquier:
> LGTM, OOI why do we check in `/vendor`? Do we do offline builds? Or is this just an OCP standard for quicker/reliable builds?

correct, dependency vendoring is the OCP standard.

PR #616, opened 2026-09-16
OCPBUGS-115550: Add 'agent-installer-postconfig' install_type

zaneb:
We will use this invoker to indicated agent-based installs that are configured after the ISO is generated (Appliance, Disconnected Assisted Installer) to distinguish them from the regular Agent-Based Installer where the full cluster configuration is embedded in the ISO.


simonpasquier (commented):
/approve
/lgtm
/verified by @simonpasquier

